PR #3, opened 2017-12-14
Disable the other tabs when edit model file or service file

guglo:
On validate config files page, when user is in the edit mode, it hides the navigation buttons. User can switch tabs while the navigation buttons are hidden.
Fixed the issue by preventing user to go to different tab while edit file is open.

### DIFF
--- a/src/pages/ValidateConfigFiles.js
+++ b/src/pages/ValidateConfigFiles.js
@@ -397,18 +397,18 @@ class ConfigPage extends BaseWizardPage {
         </div>
         <div className='wizard-content'>
           <Tabs id='configTabs' activeKey={this.state.key} onSelect={(tabKey) => {this.setState({key: tabKey});}}>
-            <Tab disabled={this.state.disableTab && this.state.key !== TAB.MODEL_FILES}
+            <Tab disabled={this.state.disableTab}
               eventKey={TAB.MODEL_FILES} title={translate('validate.tab.model')}>
               <ValidateConfigFiles disableTab={this.disableTab}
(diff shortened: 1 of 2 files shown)


src/pages/ValidateConfigFiles.js
```diff
@@ -390,15 +397,19 @@ class ConfigPage extends BaseWizardPage {
         </div>
         <div className='wizard-content'>
           <Tabs id='configTabs' activeKey={this.state.key} onSelect={(tabKey) => {this.setState({key: tabKey});}}>
-            <Tab eventKey={TAB.MODEL_FILES} title={translate('validate.tab.model')}>
-              <ValidateConfigFiles enableNextButton={this.enableNextButton} showNavButtons={this.showNavButtons}
+            <Tab disabled={this.state.disableTab && this.state.key !== TAB.MODEL_FILES}
```
jeremy-moffitt: I'm confused, why are we keeping the current tab enabled? is that for the styling? its still clickable, which causes a flicker effect... does making it disabled mess up the styling?

guglo: By default all the tabs are clickable (enabled). Logic is to  make other tabs not clickable when editing in current tab. Current tab is the one with green underline. If other tabs are clickable, when hover over the grey underlines show. If other tabs are not clickable, gray underlines don't show. disabled is the build-in attribute of Tab.

jeremy-moffitt: I found it marginally odd that the current tab is still clickable (which doesn't do anything)... its probably fine, just a little weird

guglo: @jeremy-moffitt will take a look.

PR #6, opened 2017-12-15
Fixed bugs based on UX feedback

guglo:
Then changes include: 

1. When no model is selected, model selection details pane should say "No model selected"
2. When a model is selected, model selection details pane should include a title with the model name.  3. The selected checker button should have a check . 
4. When no component is selected, the component details pane should say "No component selected"
5. When a component is selected, the information describing it should be in an info box with an info icon, The selected checker button should have a check . 
6. The model assignment accordion should not have blank space below the final entry if that entry is not expanded 
7. Add <component> links in Manage Cloud settings should be clickable on the text and icon, not just the icon
8. Check icon in wizard progress bubbles should be replaced with "done" icon form material icons library
9. Templates and Services tree should expand/collapse on text click in addition to arrow-click
10. Tooltips should have default bootstrap background (black)
11. failed deployments should use a smaller error icon (the ! looking one) 
12. While model validation is going on, the validate button should be disabled and there should be a spinning refresh icon next to "validating"
13. Separate out text instructions on Review Configuration Files page into info blocks 
14. "Import from csv" on assign servers page should be updated to be a link rather than a button

### DIFF
--- a/src/localization/bundles/en.json
+++ b/src/localization/bundles/en.json
@@ -29,6 +29,7 @@
     "logviewer.autoscroll": "Automatically scroll to the bottom",
     "no.options.available": "No options available",
     "no.component.select": "No Component Selected",
+    "no.model.select": "No Model Selected",
     "common.details": "Details",
 
     "install.intro.message.body1": "You are about to install a new cloud.",
--- a/src/localization/bundles/en.json
+++ b/src/localization/bundles/en.json
@@ -28,6 +28,8 @@
     "wizard.loading.pleasewait": "Loading... Please wait",
     "logviewer.autoscroll": "Automatically scroll to the bottom",
     "no.options.available": "No options available",
+    "no.component.select": "No Component Selected",
+    "common.details": "Details",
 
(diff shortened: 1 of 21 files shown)


src/localization/bundles/en.json
```diff
@@ -28,6 +28,8 @@
     "wizard.loading.pleasewait": "Loading... Please wait",
     "logviewer.autoscroll": "Automatically scroll to the bottom",
     "no.options.available": "No options available",
+    "no.component.select": "No Component Selected",
```
GarySmith: Per the bug report, when no model is selected, it is supposed to say "No model selected", and with this change it says "No Component Selected"

guglo: @GarySmith  thanks for the review. Ah..good catch...I just reused whatever it is there. I will make the wording change and also update the commit messages.

jeremy-moffitt:
Most minor nit-pick ever... the "refresh" icon being used to indicate that validation is on-going should not have the cursor that makes it appear clickable

PR #7, opened 2017-12-15
Show user-friendly name of selected model after selection

jgravel:
SCRD-2026 Show user-friendly name on subsequent pages after the user made a model selection. The name is showed in the upper right hand corner of the page.

Also includes a fix for a lints complaint in BaseWizardPage.js.


guglo:
@jgravel  The code looks fine. The only concern is if late we need add a page context help icon at the up right corner then model name + a ? icon, might look a little bit awkward. UX input might be needed.

PR #14, opened 2017-12-21
Change consumer attributes field to text area

jgravel:
SCRD-2241 The consumer attributes for a device group in the Disk Models tab of the Manage Cloud Settings modal was not able to accommodate different forms of consumer attributes because it was a text input field. Changed it to a textarea input field and added logic to handle the conversion between
yaml and javascript object.


jgravel:
entry-scale-ironic-multi-tenancy and entry-scale-kvm-ceph have good example of device group consumer attributes. Where did you see the error message in the console?

jeremy-moffitt:
the error was when I ran the config processor, it showed up in the main error/log section

PR #15, opened 2017-12-21
Added addresses field to add/update network

guglo:
This checkin includes:
1. Added addresses input fields to add/update network
2. Added addresses validation to validate the list like:
  1.1.2.2 - 1.1.2.3 or 1.1.2.34
  Only validate the format, don't validate if the range is reasonable, or addresses are
  within cidr.
3. Adjusted the styling for the network details section.
4. Added tooltip for cidr, vlanid and addresses to help user for network fields.
5. Mark the Server Group with * to indicate is required.

### DIFF
--- a/src/localization/bundles/en.json
+++ b/src/localization/bundles/en.json
@@ -229,6 +229,8 @@
     "network.add": "Add Network",
     "network.gateway": "Gateway IP",
     "network": "Network",
+    "network.addresses": "Addresses",
+
     "vlanid": "VLAN ID",
     "cidr": "CIDR",
     "gateway": "Gateway",
@@ -348,6 +350,7 @@
     "input.validator.networkinterface.error": "Please enter a valid network interface",
     "input.validator.vlanid.error": "Please enter a valid VLAN id (1-4094)",
     "input.validator.cidr.error": "Please enter a valid CIDR",
+    "input.validator.addresses.error": "Please enter a valid IP address or IP address range",
     "input.validator.uniquename.error": "Name exists, please specify a different name",
     "input.validator.name.spaces.error": "Name should not contain spaces",
     "input.validator.yaml.error": "Invalid YAML",
@@ -393,5 +396,9 @@
     "model.summary.role.description.OVSVAPP-ROLE" : "Describe OVSVAPP Nodes",
     "model.summary.role.description.IRONIC-COMPUTE-ROLE" : "Describe Ironic Compute Nodes",
     "model.summary.role.description.NOT_FOUND" : "Describe the fact that the customer must have created this role",
-    "model.summary.role.component.NOT_FOUND": "Custom Component Nodes"
+    "model.summary.role.component.NOT_FOUND": "Custom Component Nodes",
+
+    "tooltip.network.addresses": "IP addresses used to limit the network access. Addresses could be an IP range or an IP address. For example, 10.0.1.1-10.0.1.10 or 10.0.1.23",
+    "tooltip.network.cidr": "Network address in Classless Inter-Domain Routing notation, such as 10.0.1.0/24",
+    "tooltip.network.vlanid": "A valid VLAN id is an integer in the range of 1-4094."
 }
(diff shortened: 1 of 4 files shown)


src/localization/bundles/en.json
```diff
@@ -393,5 +396,9 @@
     "model.summary.role.description.OVSVAPP-ROLE" : "Describe OVSVAPP Nodes",
     "model.summary.role.description.IRONIC-COMPUTE-ROLE" : "Describe Ironic Compute Nodes",
     "model.summary.role.description.NOT_FOUND" : "Describe the fact that the customer must have created this role",
-    "model.summary.role.component.NOT_FOUND": "Custom Component Nodes"
+    "model.summary.role.component.NOT_FOUND": "Custom Component Nodes",
+
+    "tooltip.network.addresses": "IP addresses used to limit the network access. Addresses could be an IP range or an IP address. For example, 1.2.2.4-1.2.4.5 or 1.2.6.4",
+    "tooltip.network.cidr": "CIDR for the network in the format of 1.2.2.3/4",
```
GarySmith: ``1.2.2.3/4`` is neither a valid CIDR or good formatting example. It would be better to use text like 'Network address in Classless Inter-Domain Routing notation, such as 10.0.1.0/24'.

src/pages/ServerRoleSummary/UpdateNetworks.js
```diff
@@ -91,16 +110,55 @@ class UpdateNetworks extends Component {
...
+
   handleUpdateNetwork = () => {
     let model = this.props.model;
     for (let key in this.data) {
       if(this.data[key] === undefined || this.data[key] === '') {
         delete this.data[key];
       }
+
+      if (key === 'addresses') {
+        let cleanAddrs = this.state.addresses.filter((addr) => {
```
GarySmith: This would be little cleaner and slightly less confusing to use:
``let cleanAddrs = this.state.addresses.filter(addr => addr !== '');``

src/pages/ServerRoleSummary/UpdateNetworks.js
```diff
@@ -132,6 +190,71 @@ class UpdateNetworks extends Component {
       .sort(alphabetically);
   }
 
+  removeAddress = (idx) => {
+    this.setState(prev => {
+      let addresses = prev.addresses.slice();
+      addresses.splice(idx, 1);
+      return {addresses: addresses};
+    });
+    if(!this.state.isFormChanged) {
```
GarySmith: Instead of using a separate if, just put the ``isFormChanged: true`` in the structure returned on line 197.

src/utils/InputValidators.js
```diff
@@ -141,6 +143,32 @@ export function CidrValidator(cidr) {
...
+  // just one IPV4 address
+  if(addresses && addresses.indexOf('-') === -1) {
+    if(IPV4ADDRESS.exec(addresses.trim()) === null) {
+      retValue.isValid = false;
+      retValue.errorMsg = translate('input.validator.addresses.error');
+      return retValue;
+    }
+  }
+
+  if(addresses && addresses.indexOf('-') !== -1) { // just one range
```
GarySmith: This really needs some basic logic to cover some simple things like the first address must be less than the second address.

src/localization/bundles/en.json
guglo: ok, will fix the example.

src/pages/ServerRoleSummary/UpdateNetworks.js
guglo: ok, will update

src/pages/ServerRoleSummary/UpdateNetworks.js
guglo: ok. will update

src/utils/InputValidators.js
guglo: Ok, thanks. Will add one more basic logic based on @GarySmith's suggestion. For more complicated network logic validation, we will depend on the configure processor.

PR #17, opened 2018-01-03
Add 'Manage Subnet and Netmask' feature

jgravel:
SCRD-2213 Add the 'Manage Subnet and Netmask' button to launch a modal that allows updating of baremetal subnet and netmask


src/pages/AssignServerRoles/BaremetalSettings.js
```diff
@@ -0,0 +1,93 @@
...
+        <ActionButton clickAction={this.saveBaremetalSettings} displayLabel={translate('save')}
+          isDisabled={!this.checkSettingsChanged()}/>
+      </div>
+    );
+    return (
+      <ConfirmModal show={this.props.show} className={'manual-discover-modal'}
+        title={translate('add.server.set.network')} onHide={this.props.cancelAction} footer={footer}>
+        <div className='description-line'>{translate('add.server.set.network.description')}</div>
+        <div className='server-details-container'>
+          <ServerInputLine label={'add.server.set.network.subnet'}
```
GarySmith: It is confusing that subnet and netmask are both optional, since this would permit a user to specify a netmask without a subnet (or vice versa). I would expect them to both be required. If, on the other hand, these field are not required, it would be good to have logic to enforce that they either both be populated or both be blank.

src/pages/AssignServerRoles/BaremetalSettings.js
```diff
@@ -0,0 +1,93 @@
...
+    super(props);
+    const baremetal = this.props.model.getIn(['inputModel', 'baremetal']).toJS()
+    this.state = {
+      subnet: baremetal.subnet,
+      netmask: baremetal.netmask
+    }
+    this.origBaremetal = JSON.parse(JSON.stringify(baremetal));
+  }
+
+  handleInputChange = (e, valid, props) => {
```
GarySmith: When I put in an invalid subnet (e.g. 1.2.3.4.5), the field validator is invoked and shows red (Please enter a valid IPV4 address), but the Save button is still enabled, and it will actually save a portion of the value that I entered (which is clearly not correct). If I put in a value like 'suse', it is also permitted but seems to do nothing. The Save button should be disabled when the values are not valid

PR #21, opened 2018-01-05
Improve dev version of build

GarySmith:
Enable source maps in webpack build so that the browser can identify
which map lines in the combined javascript back to the corresponding
source file.  This enables viewing and and debugging separate source
files in the browser (instead of the minified file).  Configure the
babel transpiling so that is produces JavaScript that is runnable on
modern browsers, which all support ES2015 (aka ES6), reducing the amount
of manipulation done to the JavaScript.


GarySmith:
I forgot to mention in the commit message that I removed server.js, since this file was unused and it was confusing the issue.

PR #24, opened 2018-01-09
Ask for confirmation before exiting Manage Cloud Settings modal

jgravel:
SCRD-1950 Add a confirmation modal before exiting the Manage Cloud Settings modal when detecting changes to any of the tabs so that the user won't lose the changes accidentally. 

Also fix the logic to detect changes in the Networks tab which didn't compare the new value with the original value which caused the tab to report a change even when the new value and the original value are the same.

The change in ServerRoleSummary.js is an enhancement to expand the first group when there's no COMPUTE-ROLE group. An example of this case can be seen when you selected Entry Scale Ironic Multi-tenant Network.

### DIFF
--- a/src/pages/ServerRoleSummary/EditCloudSettings.js
+++ b/src/pages/ServerRoleSummary/EditCloudSettings.js
@@ -15,7 +15,7 @@
 import React, { Component } from 'react';
 import { translate } from '../../localization/localize.js';
 import { Tabs, Tab } from 'react-bootstrap';
-import { ConfirmModal } from '../../components/Modals.js';
+import { ConfirmModal, YesNoModal } from '../../components/Modals.js';
 import NicMappingTab from './NicMappingTab.js';
 import ServerGroupsTab from './ServerGroupsTab.js';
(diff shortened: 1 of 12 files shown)


src/pages/ServerRoleSummary/EditCloudSettings.js
```diff
@@ -35,38 +35,70 @@ class EditCloudSettings extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      key: TAB.NIC_MAPPINGS
+      key: TAB.NIC_MAPPINGS,
+      showCloseConfirmation: false
     };
+    this.dataChanged = [false, false, false, false, false];
```
jeremy-moffitt: This is a little bit painful, as we'll have to remember to expand it if we add additional tabs

jgravel: I'm all ears if you have a better idea.

jgravel:
What template did you use? I forgot about the bonded network cases and didn't see them from the templates I was testing with. Good catch.

PR #37, opened 2018-01-20
Feat: show other useful information in details of the servers managed by SUMa

mbologna:


src/pages/AssignServerRoles/ModelServerDetails.js
```diff
@@ -45,6 +45,8 @@ class ModelServerDetails extends Component {
           {this.renderTextLine('server.name.prompt', this.props.data.name)}
           {this.renderTextLine('server.role.prompt', this.props.data.role)}
           {this.renderTextLine('server.ip.prompt', this.props.data['ip-addr'])}
+          {this.renderTextLine('server.ip.prompt', this.props.data['release'])}
```
GarySmith: This page shows parameters that have been manually entered, i.e. not the fields that were discovered when talking to SUMA or oneview. There is no place in the UI where these are manually entered, so they should not appear here. Note also that they use they wrong label here (3 fields on the page are now labelled as IP address)

guglo:
Chatted with @mbologna  think os release and kernel might be better stay in the suma details view.

mbologna:
Modified: upon discussion we decided to show detailed information such as running os and kernel etc. on the details, since not all providers (e.g. OneView) might have that information.

NOTE: this PR depends on https://github.com/ArdanaCLM/ardana-installer-ui/pull/37

PR #38, opened 2018-01-22
SCRD-2576 Show meaningful string when new model names are added

GarySmith:
When a previously-unknown model name is encountered, render it
in a reasonable way.  Also renamed the fixed translatable string names
from values like 'model.picker.foo' to 'model.name.foo' since these
strings are in many places in the UI beside the model picker page.


GarySmith:
Good idea, Gloria. I will add that

guglo:
Looks good. Thanks!

PR #44, opened 2018-01-29
Fixed the status problem for cloud deploy process

guglo:
This checkin includes:
1. Fixed the status problem where the status was not updated correctly
when there are some playbooks included in step but acutally they might
not get run based on user settings,  for example monasca playbooks.
2. Also added a condition where to detect only need one playbook gets run
to decide if the step completes. For example site.yml or dayzero-site.yml.
3. Updated to have ardana-status.yml for the health step

### DIFF
--- a/src/components/PlaybookProcess.js
+++ b/src/components/PlaybookProcess.js
@@ -118,6 +118,13 @@ class PlaybookProgress extends Component {
     return retClass;
   }
 
+  getStepCompletedPlaybooks = (step) => {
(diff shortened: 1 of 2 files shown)


guglo:
To test: 

Update ardana service to https://gerrit.suse.provo.cloud/#/c/2990/
In order to mimic the situation where ceph-deploy.yml event is not there...you can also temp update the site.log to remove ceph-deploy.yml start and stop event.

PR #47, opened 2018-01-30
Updated copyright 2017 to be 2018

guglo:


jeremy-moffitt:
as with the ardana-installer-server pull request, I may have miscommunicated the need here.. 
If theres a single date of 2017, it should become 2017-2018
if theres an existing range that ends in 2017 (say 2015-2017), then the end-year should be updated to 2018... i.e. 2015-2018

PR #49, opened 2018-02-01
Show the log during file editing if validation failed

jeremy-moffitt:
Displays the config processor output when editing a yml file if the config processor failed validation


guglo:
@jeremy-moffitt  maybe when user clicks validate instead of allow them to click other tab to update file...maybe disable the tab when in the process of validating?

guglo:
Everything looks good. Just one nit, can you disable the next button while it is validating as well? Thanks.

jeremy-moffitt:
Uploaded new patch set with the back button disabled during validation.... next was already disabled until validation had succeeded

PR #55, opened 2018-02-02
Add confirmation modal when changing cloud model

jgravel:
SCRD-2470 Added a confirmation modal when user selected a different cloud model than before. This is a generic warning to the user and does not actually keep track of any change that may happen after the user selected a cloud model for the first time.


guglo:
@jgravel  for some reason, the confirmation doesn't come out for me which switch model.

guglo:
@jeremy-moffitt thanks.

PR #56, opened 2018-02-05
WIP Upated to make ID editable

ghost:
This checkin includes:

1. Removed name field which is not model field.
2. Added uid internal field for update or remove.
3. If duplicated items in right side table, turn red.
4. ID field is now editable. It should be unique.

Note:
Plan to deal with the following issue separately:
In server details view, hostname, IP and MAC address could be out of sync
if server details get changed at the right side and drag back to left side.
Adding Server Manually allow duplicates.

It goes with https://github.com/ArdanaCLM/ardana-installer-server/pull/16


guglo:
Made another change without using uid. Abandon this.

PR #58, opened 2018-02-06
Make the id editable

ghost:
This checkin includes:

1. Removed name field in edit server modal and model
2. Made the id editable in edit server modal
3. Added unique id validation to make sure no duplicate id updated/added
4. Fixed to not allow  duplicate of server added manually
5. Fixed problem where the last valid input can turn on the add more or save
button while the previous invalid inputs are ingnored.
6. Cleanup for unused cpu and ram item in the server data

This goes with backend change:
https://github.com/ArdanaCLM/ardana-installer-server/pull/17

Note:
To make this change smaller, will address import duplicates in csv seperately
since it has a different bug item.
Also if import csv directly without editing, it will have empty field like ilo_user,
ilo_password...which can cause problems for config processor. Will addrss this seperately
as well. .


guglo:
kid of messed up with comments when rebase and push -f ...will have a new pull request since no one has looked at it yet.

PR #60, opened 2018-02-07
Make the id editable

ghost:
This checkin includes:

    1. Removed name field in edit server modal and model
    2. Made the id editable in edit server modal
    3. Added unique id validation to make sure no duplicate id updated/added
    4. Fixed to not allow  duplicate of server added manually
    5. Fixed problem where the last valid input can turn on the add more or save
    button while the previous invalid inputs are ingnored.
    6. Cleanup for unused cpu and ram item in the server data
    7. Fixed issue when user updates in server role summary but the leftside
    servers are not updated accordingly.
    8. Fixed issue where the accordion table has part of gray and part of white
    when overflow

    This goes with backend change:
    https://github.com/ArdanaCLM/ardana-installer-server/pull/17

    Note:
    To make this change smaller, will address existing problems separately
    1. import duplicates in csv
    2. if import csv directly without editing, or add server without editing,
    or auto discovered server edited, but drag back to left side and drag to right again,
    it will have empty field like ilo_user, ilo_password...which can cause problems
    for config processor.


guglo:
Will use uid approach.

PR #66, opened 2018-02-13
Added confirmation before import servers

ghost:


src/localization/bundles/en.json
```diff
@@ -221,7 +221,7 @@
     "server.secure": "Verify SSL certificate",
     "server.import.update.error": "Failed to save updated imported server {0}",
     "server.import.add.error": "Failed to save new imported servers",
-
+    "server.import.server.confirm": "The imported servers will overwrite previously imported servers if they have the same ids.\n If imported servers happen to have the same ids as the manually added servers or discovered servers, servers with duplicate ids could show up. If drag the servers with the same ids to assign roles, servers with duplicate ids will be in red. You can edit the server to update the id.\n\n Do you still want to import servers ? ",
```
jeremy-moffitt: "If drag the servers with the same ids to assign roles" needs to be wordsmithed a bit....

jeremy-moffitt (changes requested):

src/localization/bundles/en.json
guglo: Ok...will work on the wordings

jeremy-moffitt (approved):

PR #67, opened 2018-02-13
Add guidance on changing number of nodes

jgravel:
Bugzilla 1080184 - added an info banner on the Cloud Model Summary page when user selects a mandatory component that has 3 nodes


jgravel:
Thanks @jeremy-moffitt for the comment. I made an update so that the initial node number was used to decide whether or not to show the banner. Please take a look again.

PR #70, opened 2018-02-21
Remove storage type from filter bar

jgravel:
SCRD-2193  Updated template names since the input model no longer has references to SES. Therefore having the storage type filter box with one available value 'Swift' does not make sense. Commented out one line of code for now, in case they want to resurrect SES in the future.

Depends on this backend change: https://gerrit.suse.provo.cloud/#/c/3335 which is going through the gate at the moment


jgravel:
The backend change was merged into master. Please feel free to test it with this UI change.

PR #71, opened 2018-02-23
Remove SUSE from default page title

jgravel:
Bugzilla 1082464 - replaced "SUSE OpenStack Cloud Deployer" with "OpenStack Cloud Deployer" in the page title

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
-    <title>SUSE OpenStack Cloud Deployer</title>
+    <title>OpenStack Cloud Deployer</title>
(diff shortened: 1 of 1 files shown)


rsalevsky:
Sorry but this is just wrong for multiple reasons.
1. This is not a fix but more a workaround; we must have SUSE in the title as this is the product name.
The real fix would be to to replace SUSE with HPE in the branding patch we have in the package.
2. The commit message must have the Bugzilla number in the commit header so for example `My commit message (bsc#123456)`. There was a mail on cloud-devel and this is also how gerrit expects/forces it; in this critical phase there should be no commit without Jira or Bugzilla reference in the commit header.

So please revert this change and fix it properly.

jeremy-moffitt:
@rsalevsky the title does have SUSE in it , the javascript code updates from the default in html to the branded version (it will up to SUSE in the SUSE version). The issue is that there is a split second where the default HTML title is shown before the javascript can update it. It's less confusing for customers who are watching carefully to see it change from "OpenStack..." to "HPE OpenStack.." or "SUSE OpenStack..." than it is for it to change from "SUSE OpenStack..." to "HPE OpenStack..." . Rest assured that the fully loaded product will have SUSE in the title.

PR #73, opened 2018-02-23
Add Edit Server modal for servers added manually (bsc#1080193)

jgravel:
Added the Edit Server modal for servers added manually. The work includes refactoring out the existing code that handles servers added manually to be it's own module and reusing it for the edit modal.

Also fixed the View Details modal so that it's look and feel is consistent with the Add and Edit modals.

The changes in CloudModalSummary and CsvImporter are fixes for lints complaints.

### DIFF
--- a/src/components/ServerTable.js
+++ b/src/components/ServerTable.js
@@ -53,9 +53,16 @@ class ServerTable extends Component {
       });
 
     // push an empty header to hold show detail icon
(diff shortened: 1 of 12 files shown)


jgravel:
I verified that the behavior that @jeremy-moffitt found was an existing behavior in master. I'll see if I can include the fix in this change.

PR #74, opened 2018-02-28
Fixed to handle end of last playbook when no end event (bsc#1083243)

ghost:


jeremy-moffitt (approved):

jeremy-moffitt:
tried it out locally and works fine... won't know about a full QA/QE environment until its in the build

PR #75, opened 2018-03-01
Fix warning modal when select template for the first time (bsc#1082902)

jgravel:
The propsModelName variable used to return an undefined but now returns a null. This causes the warning modal to come up when the template was selected for the first time which is not a correct behavior. Changed from '===' to '==' which takes the null value into account in that check.


GarySmith:
That's why I hate === and the jslint rules that try to make you use it all of the time. Sometimes it just makes things work much better and easier.

PR #81, opened 2018-03-13
Fixed goback issue for ConfigDeployProgress page (bsc#1085146)

ghost:
When the deploy progress page has errors, user can click back button
to go back to the previous page. If user checked wipedisk last time,
it ran dayzero-site.yml last time. Click back button will reset
wipedisk to false, while the the global playbook status didin't update
the status to use site.yml. Updated to set the playbook status
correctly.

### DIFF
--- a/src/pages/CloudDeployProgress.js
+++ b/src/pages/CloudDeployProgress.js
@@ -78,6 +78,8 @@ class CloudDeployProgress extends BaseWizardPage {
   constructor(props) {
     super(props);
 
+    this.playbooks = [];
+
     this.state = {
       overallStatus: STATUS.UNKNOWN // overall status of entire playbook
     };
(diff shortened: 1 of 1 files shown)


guglo:
To test in dev env

Need to cause error to enable the back button, update PlaybookProcess.js line 305 log to be log1 or 322  event to be event 1. This will cause error when fresh the deploy page. 

Check wipe disk first time on config page, validate,  click next to get into deploy page, let it run and finish. 
Check the progress.json, it should indicate dayzero-site in playbookStatus and wipeDisks is true.
Once deploy is done, refresh the deploy page, you will get errors and the back button will be enabled because of the force error of PlaybookProgress.js.  Click the back button to go back to config page. Validate the model and click next again to start another deploy. check the progress.json, you will see this time wipeDisks is false and site in playbookStatus.

src/pages/CloudDeployProgress.js
```diff
@@ -95,14 +97,9 @@ class CloudDeployProgress extends BaseWizardPage {
...
-      playStatus.forEach((play, idx) => {
-        // remove the exist one
-        if (play.name === PRE_DEPLOYMENT_PLAYBOOK ||
-          play.name === DAYZERO_SITE_PLAYBOOK || play.name === SITE_PLAYBOOK) {
-          play.playId = '';
-          play.status = '';
-        }
+      // remove playbook status for any playbook in this.playbooks
+      let playStatus = this.props.playbookStatus.filter(status => {
+        return this.playbooks.indexOf(status.name) === -1;
```
GarySmith: While this works, it is more intuitive and correct to use [includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) than comparing indexOf() with -1. For example:
```
let playStatus = this.props.playbookStatus.filter(status => ! this.playbooks.contains(status.name));
```

GarySmith (changes requested):

GarySmith (approved):

PR #84, opened 2018-03-21
Fixed blank page on firefox (bsc#1086071)

ghost:
Firefox doesn't support finally in version < 58. Use then to be safe.


GarySmith (approved):
Works for me

PR #87, opened 2018-03-28
Add Manage Server Role modal

jgravel:
SCRD-1607 Added Manage Server Role modal to the Cloud Model Summary page. The modal contains 3 tabs: Server Roles, Disk Models, Interface Models. Users can use this modal to add, edit, and remove server roles.

Also fixed a lint complaint in PlaybookProcess.js.


GarySmith:
Closing this PR so that we do not have to keep looking past it. The branch is still being retained, so we can open a new PR with this same code if we want to move forward with it.

GarySmith:
The code that was in this branch, which was a single commit, has been attached as a patch file to the jira item (https://jira.prv.suse.net/browse/SCRD-1607), and the branch has been removed. That patch will be a good starting point should this change be worked on again.

PR #89, opened 2018-04-03
Scrd 2374 cancel playbook

GarySmith:


jgravel:
I tested this patch again (with Gary's guidance). The confirmation modal came up correctly when I cancelled the deploy. But before I could click Yes or No in the modal, the failure message 'Playbook cancelled' already appeared in the text box. The message should not appear until there's a Yes or No answer on the modal.

GarySmith:
Should be working correctly now. Also fixed an existing related problem where the form could end up displaying both an error pane and a log pane, when it was never intended to do so (it was apparently assumed that any errors and their messages would preclude the playbook to begin and start logging)

PR #92, opened 2018-04-05
Fixed IE blank screen issue (bsc#1087181)

ghost:
This checkin includes:
1. Added babel-ployfill to deal with Object.assign, Array.includes issues with IE
2. Added whatwg-fetch to deal with fecth issue with IE
3. Used  window.navigator.language instead of window.navigator.languages to handle en-US, en, ja-JP, ja and zh
4. Added drag and drop fix for IE
5. Fixed a sytling issue in IE for the search bar in assign server role page

### DIFF
--- a/src/localization/localize.js
+++ b/src/localization/localize.js
@@ -36,14 +36,12 @@ for (var i = 0; i < supportedLangs.length; i++) {
 // window.navigator.language is the language that user sets at the top of
 // installed and enabled languages like en-US, en, ja-JP, ja, zh-CN, zh
 // since we have language without country like en or ja as supported languages,
-// need to find a matching or kind of matching one to set on strings
+// need to find a exact match or match the language part
 var strings = new LocalizedStrings(catalog);
 
 var findLang =
-  supportedLangs.find(
-    sLang => window.navigator.language === sLang ||
-    window.navigator.language.startsWith(sLang)
-  );
+  supportedLangs.find(sLang => window.navigator.language === sLang) ||
+  supportedLangs.find(sLang => window.navigator.language.substring(0, 2) === sLang.substring(0, 2));
 
 if(findLang) {
   strings.setLanguage(findLang);
--- a/src/localization/localize.js
+++ b/src/localization/localize.js
@@ -14,6 +14,7 @@
 **/
 import LocalizedStrings from 'react-localization';
 
+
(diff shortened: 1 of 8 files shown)


src/localization/localize.js
```diff
@@ -40,8 +40,10 @@ for (var i = 0; i < supportedLangs.length; i++) {
 var strings = new LocalizedStrings(catalog);
 
 var findLang =
-  supportedLangs.find(sLang => window.navigator.language === sLang) ||
-  supportedLangs.find(sLang => window.navigator.language.startsWith(sLang));
+  supportedLangs.find(
+    sLang => window.navigator.language === sLang ||
+    window.navigator.language.startsWith(sLang)
+  );
```
GarySmith: I think this change is not correct, but the earlier one isn't quite either. I think what you want to do is that if you have an exact match, use it. But if there are no exact matches, then use any where the first two characters match. Then fallback to English.
In this version, if we supported zh-cn and zh-hk and the browser is set to zn-hk, it would return 'zh-cn'.
In the earlier version (which is also not quite right), if we supported zh-cn and zh-hk and the browser was set to zh, it would return 'en'.
Perhaps logic like:
```
  if supportedLangs.find(slang => window.navigator.language === sLang) {
    // Use an exact match if possible
    strings.setLanguage(sLang);
  } else {
    // Use any string whose language code matches, otherwise default to 'en'
    const match = supportedLangs.find(slang => window.navigator.language.substring(0,2) === sLang.substring(0,2))
    strings.setLanguage(match || 'en');
  }
```

GarySmith (changes requested):

GarySmith:
Other than that, the change seemed fine. It worked fine under chrome still, but I don't have MS browsers to test with.

jeremy-moffitt (changes requested):
I think something in the DnD code was missed... dragging between roles is broken with this changeset (it works to drag right to left or left to right, but not up and down the list of roles)

jeremy-moffitt:
as a side-note, it looks like both polyfill and whatwg-fetch were already being pulled into node_modules, so I dont *think* we'll need to rebuild the IBS copy of our node_modules tarball for this to work...

jeremy-moffitt (approved):
I cant reproduce the issue, changing back to approve

GarySmith (approved):

PR #95, opened 2018-05-18
InstallUI redirect if ardana-service requires auth (bsc#1090278)

jeremy-moffitt:
After a deployment is successful, the ardana-service is in a mode where it requires authentication to operate. This causes the installui to get a 401 error when making most calla to the ardana-service rest service. When this happens, the UI will now redirect to the deployment summary. In the future (when the day0 UI is also used for day2 operations post-install), it will instead redirect to the login screen

### DIFF
--- a/src/Deployer.js
+++ b/src/Deployer.js
@@ -19,6 +19,8 @@ import { pages } from './utils/WizardDefaults.js';
 
 class Deployer extends Component {
   render() {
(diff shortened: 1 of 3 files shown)


jeremy-moffitt:
obsoleted by https://github.com/ArdanaCLM/ardana-installer-ui/pull/100

PR #97, opened 2018-05-25
Refactor ServerUtils inputs to be generic (SCRD-3292)

jeremy-moffitt:
Technical debt clean-up of some inputs that were in ServerUtils.js, but are actually used in a variety of pages not tied directly to servers. Moved each component into its own file and removed "server" references. Updated the files that use these components.


src/components/ServerUtils.js
```diff
@@ -18,8 +18,8 @@ import '../Deployer.css';
 import { translate } from '../localization/localize.js';
 import ServerTable from './ServerTable.js';
 import { isRoleAssignmentValid } from '../utils/ModelUtils.js';
-
-const TRUTHY = [true, 'true', 'True', 'yes', 'Yes'];
+import { LabeledDropdown } from './LabeledDropdown.js';
+import { ValidatingInput } from './ValidatingInput.js';
 
 export class SearchBar extends Component {
```
guglo: LabeledDropdown and ValidatingInput are not sued in this file

PR #100, opened 2018-06-01
Add keystone authorization UI

GarySmith:
Create modal dialog to prompt for credentials, and verify them against
keystone via the ardana-service.  Create functions and classes for
managing token information with cookies and injecting it into subsequent
REST calls.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "babel-polyfill": "^6.26.0",
     "better-react-spinkit": "^2.0.0",
     "font-awesome": "^4.7.0",
+    "history": "^4.7.2",
     "immutable": "^3.8.1",
     "js-yaml": "^3.10.0",
     "lodash": "^4.17.4",
--- a/src/Deployer.js
+++ b/src/Deployer.js
@@ -19,28 +19,31 @@ import { pages } from './utils/WizardDefaults.js';
 import { HashRouter as Router, Switch } from 'react-router-dom';
 import Route from 'react-router-dom/Route';
 import { translate } from './localization/localize.js';
+import LoginPage from './pages/Login.js';
 
 class Deployer extends Component {
   render() {
+    //TODO - wrap the InstallWizard with a component that varies the wizard by task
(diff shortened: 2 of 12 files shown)


GarySmith:
Note that approach uses a modal dialog box for prompting for credentials, but modifying the whole routing approach may require adjusting this (see https://github.com/ArdanaCLM/ardana-installer-ui/pull/99)

GarySmith:
This version incorporates the routing prototype from pull request #100 . Two things could still be improved, but are probably best handled in a separate change:
1. When initially navigating to the UI, the first page will briefly be displayed and then redirected to the login page. (fixing this would require some login in the wizard at startup to avoid this)
2. If a valid token times out when the user is inactive, this will not be detected until the user tries to perform some operation against the ardana-service. Instead it should be more like a banking website that automatically jumps to the login page when the time has elapsed even without any activity.

GarySmith:
One additional note: this change introduces a new dependency, so the IBS job will need to be updated (to include a new node_modules tarball) with the new dependency. Failure to do so should not cause the build to fail, but if a user *were* to try to use the app in a day 2 scenario without this, then they would get an error in the browser

jeremy-moffitt:
question about:
> When initially navigating to the UI, the first page will briefly be displayed and then redirected to the login page. (fixing this would require some login in the wizard at startup to avoid this)
could this be solved by displaying nothing until we check for the token?

jeremy-moffitt:
I'm a bit confused, when I try to reproduce the problem described above (needing to click login twice) its not happening anymore... I'm not sure what changed

GarySmith:
> could this be solved by displaying nothing until we check for the token?

I think so, and that's what I intend to do. I just didn't want this change to wait for me to get that implemented.

GarySmith:
> I'm a bit confused, when I try to reproduce the problem described above (needing to click login twice) its not happening anymore... I'm not sure what changed

Same here. I was able to reproduce that for a time, but it seems to work flawlessly now. Strange.

PR #103, opened 2018-06-14
SCRD-3672 Fix id alias server-id in field_defs

mserigithub:
Fixing CsvImporter.js to allow server_id as a valid column name for importing server id.

### DIFF
--- a/src/utils/CsvImporter.js
+++ b/src/utils/CsvImporter.js
@@ -32,7 +32,7 @@ export function importCSV(file, restrictedValues, callback) {
   const field_defs  = {
     'id': {
       unique: true,
-      aliases: [ 'server_id', 'id' ],
+      aliases: [ 'server-id', 'id' ],
       required: true
     },
     'ip-addr': {


src/utils/CsvImporter.js
```diff
@@ -32,7 +32,7 @@ export function importCSV(file, restrictedValues, callback) {
   const field_defs  = {
     'id': {
       unique: true,
-      aliases: [ 'server_id', 'id' ],
+      aliases: [ 'server-id', 'id' ],
```
jeremy-moffitt: should this include both server_id and server-id ?  @GarySmith ?

guglo: aliases include the normalized key which has the - , it is consistent with the patterns for all other keys.

jeremy-moffitt: thanks for clarifying!

GarySmith:
Thanks for fixing this, Mohan

PR #105, opened 2018-06-21
SCRD-3458 Create menu system

GarySmith:
Create navigational menu system in the style of the EOS Design System [1].
Uses react-router to create a route to each of the main menu items and
sub-items.  This version supports only a fixed, static menu structure
in order to centralize the logic for routing and menu processing.
If we later need to perform dynamic routing (and dynamic menus), then
this can be introduced but may complicate the logic a bit.

### DIFF
--- a/src/components/NavMenu.js
+++ b/src/components/NavMenu.js
@@ -1,0 +1,106 @@
+// (c) Copyright 2018 SUSE LLC
+/**
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+**/
+
+import React, { Component } from 'react';
+import { HashRouter as Router, Link } from 'react-router-dom';
+import Route from 'react-router-dom/Route';
+import { translate } from '../localization/localize.js';
+
+class NavMenu extends Component {
+
+  render()
+  {
+    // Build up the left bar items. Using Route enables us to highlight
+    // the currently chosen item, i.e. the one that matches the current URL
+    const leftBarItems = this.props.routes.map((e, index) => (
+      <Route
+        key={index}
+        path={e.slug}
+        children={({ match }) => (
+          <li >
+            <Link className={match ? 'active' : ''} to={e.slug}>{e.name}</Link>
+          </li>)}
+      />
+    ));
+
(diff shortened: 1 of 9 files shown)


GarySmith:
Note that to test this you will need to run the ardana service in secured mode (integrated with keystone). Recall that this is easy to setup with docker and https://github.com/GarySmith/docker-keystone.

jack-likes-to-code:
- The login page could use a spinner since it takes a little bit of time to login.
- The line under "SUSE Openstack Cloud 8" is too short, does not reach all the way right side of the menu.  It is the same length as the highlighted menu selection
- The menu selection leaves a strange border around the selected menu item when the cursor leaves the item in Firefox.  Looks ok in Chrome.  It might be my Firefox with my extensions, so please double-check.

src/styles/components/navmenu.less
```diff
@@ -0,0 +1,122 @@
...
+    transition: all ease-in 0.25s;
+    white-space: nowrap;
+
+  }
+}
+
+.main-window {
+  margin-left: @main-menu-width;
+  height: 100%;
+  align-items: center;
```
guglo: @GarySmith  I am trying with my servers page,  it ends up a narrow column in the middle. Remove this line seems to fix my problem. However my pages are still under development.

src/components/NavMenu.js
```diff
@@ -0,0 +1,106 @@
...
+        );
+      });
+    });
+
+    return(
+      <Router>
+        {/* Router requires a single child, so surround everything in a div */}
+        <div>
+          <aside className="main-menu">
+            <header>{translate('openstack.cloud.deployer.title.version')}</header>
```
guglo: @GarySmith For this kind of heading , SUSE usually uses an image file. For example I used the one Cythia created for me for Horizon SUSE theme https://github.com/SUSE/openstack-dashboard-theme-SUSE/tree/master/themes/suse/img/logo.svg . You might want to check.

src/styles/components/navmenu.less
GarySmith: I noticed that on the page I was doing, too. I have a simple fix for that.

src/components/NavMenu.js
GarySmith: Upon further inspection, I would prefer to leave this is a string unless we need to change it because 1) it is a text string which is much easier to deal with as text than as SVG, 2) the EOS prototype from the UX team uses text (in addition to a png logo).

GarySmith:
@guglo I didn't see your comment earlier about landing on the menu and making one of them selected by default. I think that's a good idea, but since all of the menu items are kind of bogus right now, I'm not sure that's of much value yet. Perhaps we can address that as a separate change.

src/components/NavMenu.js
guglo: @GarySmith  that is fine. I just want to point it out something to double check. We can always to deal with that later.

guglo:
@GarySmith  thanks for the fix. It is fine to implement the default landing page in the separate change.

GarySmith:
Merged as this was approved

PR #106, opened 2018-06-28
SCRD-3483 Create speedometer graph component

GarySmith:
Create a speedemeter graph component that is modelled after the
"horseshoe" graph in the Operations Console.  Render the various SVG
components with a variety of configurable properties that enable
fine-tuning its appearance.  Create a test page for interactively
exercising the various properties of the speedometer.  Added the
test page into the prototype menu system.


GarySmith:
FYI, the sample page for the speedometer adds a border around the component just to clearly identify its boundary, and this it is not part of the component.

PR #109, opened 2018-07-02
Address npm dependency warnings

GarySmith:
Update dependencies to address the warnings emitted by npm about peer
dependencies not being found.


GarySmith:
Good catch, Jeremy. Thanks

PR #110, opened 2018-07-03
Implemented framework for replacing servers

ghost:
1. Added servers landing page to show all the servers. Made it
interact with the menu.
2. Added context menu for the page and updated context menu
to be functioning
3. Updated Install wizard to deal with update page and refresh
update page whilt it is still in progress
4. Added replace server modal
5. Moved server details component to the components dir
6. Moved CollapsibleTable to the components dir
7. Added 3 placeholder pages to make sure the flow works

Note:
This is not fully tested, only tested a happy path.
There are lots of TODOs need to be resolved.

To test:

1. Test without available servers
you can just enter any ilo inputs

2. Test with available servers
need to go to day0 to get some manual servers
or discovered servers, they will  show up in the dropdown
list in replace detail modal


src/pages/UpdateServers.js
```diff
@@ -0,0 +1,198 @@
...
+          <div className='buttonBox'>
+            <div className='btn-row'>
+              <ActionButton type='default'
+                displayLabel={translate('collapse.all')} clickAction={() => this.collapseAll()} />
+              <ActionButton type='default'
+                displayLabel={translate('expand.all')} clickAction={() => this.expandAll()} />
+            </div>
+          </div>
+        </div>
+        <div className='wizard-content'>
```
jeremy-moffitt: one thing about using the wizard css classes is that its causing an overly large header and clips the scrolling awkwardly...  theres a couple of solutions, one that I tried successfully was to add another css class here that sets the height back to auto... it might make more sense to just not have things that arent part of the wizard be under "wizard-page" and "wizard-content" though if that creates more css issues than it solves it might not be worth it

jeremy-moffitt: hmm it does mess with the margins and whatnot to remove those :-/ will push a patchset with the css class approach

src/styles/pages/ServerRoleSummary.less
```diff
@@ -305,58 +305,4 @@
...
-      float: right;
-      padding-right: 1em;
-      width: 100%;
-      text-align: end;
-    }
-    .expand-collapse-icon {
-      .material-icons {
-        padding-left: 0.3em;
-        vertical-align: middle;
-        color: @box-border-color;
```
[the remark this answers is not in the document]
jeremy-moffitt: uploaded a patchset to put this styling back in place

src/InstallWizard.js
```diff
@@ -63,31 +62,83 @@ class InstallWizard extends Component {
...
 
     // Indicate which of the state variables will be persisted to, and loaded from, the progress API
     this.persistedStateVars = [
       'currentStep', 'steps', 'playbookStatus', 'connectionInfo', 'deployConfig'
     ];
+
+    if(this.state.currentMenuName) {
+      let extraVars = ['currentMenuName', 'currentOperation', 'operationProps'];
+      this.persistedStateVars = this.persistedStateVars.concat(extraVars);
+    }
```
jeremy-moffitt: I'm a little confused about what this is doing, and what the impact may be of the persistedStateVars list being variable depending on other conditions.... so we're sometimes persisting 3 extra vars and sometimes not....

src/InstallWizard.js
```diff
@@ -178,7 +241,22 @@ class InstallWizard extends Component {
...
+      props.cancelUpdateProcess = this.cancelUpdate;
+    }
+
+    //don't have any progress steps
+    if (this.state.currentStep === undefined) {
+      return React.createElement(this.props.menuComponent, props);
+    }
+    else {
+      //if it is update, it will use dynamically assembled this.state.pages when startUpdate is
+      //triggered, if it is install, it will use the static this.props.pages
```
jeremy-moffitt: if what is update? is this basically splitting between "install" and "configure" modes?

src/InstallWizard.js
```diff
@@ -274,6 +352,50 @@ class InstallWizard extends Component {
     }
   }
 
+  //this is triggered when start the update processes
+  //for example click repalce button on Replace server modal
+  startUpdate = (operation, pages, theProps) => {
```
jeremy-moffitt: its a little weird that this stuff is in the InstallWizard.js , I guess it allows for reusing the wizard functionality when doing a multi-step process as a result of a user choice, so its probably ok

src/components/CollapsibleTable.js
```diff
@@ -60,6 +66,68 @@ class CollapsibleTable extends Component {
     this.setState({showServerDetailsModal: true, activeRowData: rowData});
   }
 
+  handleShowMenuServerDetails = () => {
```
jeremy-moffitt: it occurs to me reading this that this component is very server list specific, but thats a pre-existing condition and not something introduced here

src/components/PlaybookProcess.js
```diff
@@ -506,7 +506,10 @@ class PlaybookProgress extends Component {
 
   componentDidMount() {
     this.globalPlaybookStatus = this.getGlobalPlaybookStatus();
-    this.processPlaybooks();
+    //TODO don't start process if passed in how to display commit error???
```
jeremy-moffitt: this comment isnt really clear

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
+// (c) Copyright 2017-2018 SUSE LLC
```
jeremy-moffitt: new file copyrights can just be 2018

src/pages/CloudDeployProgress.js
```diff
@@ -162,7 +162,7 @@ class CloudDeployProgress extends BaseWizardPage {
           <PlaybookProgress
             updatePageStatus = {this.updatePageStatus} updateGlobalState = {this.props.updateGlobalState}
             playbookStatus = {this.props.playbookStatus}
-            steps = {PLAYBOOK_STEPS} deployConfig = {this.props.deployConfig}
```
jeremy-moffitt: is deployConfig no longer used?

src/pages/ServerRoleSummary.js
```diff
@@ -28,10 +28,6 @@ class ServerRoleSummary extends BaseWizardPage {
     super(props);
 
     this.checkInputs = ['nic-mapping', 'server-group'];
-    this.state = {
-      autoServers: undefined,
-      manualServers: undefined
-    };
```
jeremy-moffitt: these are still referenced, is this just cleanup since they'll default to undefined?

src/styles/components/collapsibletable.less
```diff
@@ -0,0 +1,65 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+**/
+
+.collapsible-table {
+  .group-row {
+    border-bottom: 1px solid @box-border-color;
+    td {
+      background: linear-gradient(to bottom, @table-header-gradient-color 0%, @table-header-color 100%);
```
jeremy-moffitt: this is cool :)

src/styles/components/contextmenu.less
```diff
@@ -17,12 +17,9 @@
 .context-menu-container {
   position: absolute;
   z-index: 999;
-  width: 120px;
+  width: 103px;
```
jeremy-moffitt: wondering if this will be enough space if we get some longer action names...

src/utils/DynamicPagesFinder.js
```diff
@@ -0,0 +1,35 @@
...
+* limitations under the License.
+**/
+
+import {UpdateServerPages} from "../pages/ReplaceServer/UpdateServerPages";
+
+const PAGES_HOLDERS = {
+  'servers_ReplaceServer': UpdateServerPages //<menu>_<operation> : <all the possible pages defined>
+};
+
+export function getPages(menuName, operationName, steps) {
```
jeremy-moffitt: its not at all clear what this class or function are doing... needs comments

jeremy-moffitt (changes requested):
generally speaking looks good... theres some comments I made outside of the "review" (yes, I should've just started a review) that need a look... please take a look at those

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+  updateAutoFillFormValidity = (newJSData) => {
+    for (let key in newJSData) {
+      if(newJSData[key]) {
+        this.allInputsStatus[key] = INPUT_STATUS.VALID;
+      }
+      else {
+        this.allInputsStatus[key] = INPUT_STATUS.INVALID;
+      }
+    }
+    this.setState({isFormValid: this.isFormTextInputValid()});
```
GarySmith: remove

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  handleOsPasswordChange = (e) => {
+    const password = e.target.value;
+    this.setState({osInstallPassword: password}, () => {
+      this.setState({isFormValid: this.isFormTextInputValid()});
```
GarySmith: Remove. 
Not only is this unnecessary (as mentioned above), this code is actually wrong. this.setState executes in a promise, and when that promise returns, then rendering can begin. But the callback here is another call to setState which immediately returns a promise, so `isFormValid` is likely to not be set until *after* rendering, which is wrong.

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+    });
+  }
+
+  handleInstallOsCheck = () => {
+    this.setState({
+      isInstallOsChecked: !this.state.isInstallOsChecked }, () => {
+      this.setState({isFormValid: this.isFormTextInputValid()});
```
GarySmith: Ditto above. Remove this

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+    );
+  }
+
+  renderFooter() {
+    return (
+      <div className='btn-row input-button-container'>
+        <ActionButton type='default'
+          clickAction={this.handleCancel} displayLabel={translate('cancel')}/>
+        <ActionButton
+          isDisabled={!this.state.isFormValid}
```
GarySmith: Replace the reference to `this.state.isFormValid` with a call to `this.isFormTextInputValid`

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+import { genUID, getAvailableServerIds, maskPassword } from '../utils/ModelUtils.js';
+import HelpText from '../components/HelpText.js';
+import { Map, fromJS } from 'immutable';
+import {fetchJson} from "../utils/RestUtils";
+
+class ReplaceServerDetails extends Component {
+  constructor(props) {
+    super(props);
+
+    this.allInputsStatus = {
```
GarySmith: These status fields should be part of state, since they directly affect what is being rendered

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+      allAvailableServers = this.props.autoServers;
+    }
+    if(this.props.manualServers && this.props.manualServers.length > 0) {
+      allAvailableServers = allAvailableServers.concat(this.props.manualServers);
+    }
+
+    let theServer = allAvailableServers.find(server => server.id === serverId);
+    let newJSData = {};
+    if(theServer) {
+      newJSData['ilo-ip'] = theServer['ilo-ip'] ? theServer['ilo-ip'] : '';
```
GarySmith: While this works, it is more straightforward to use:
```
newJSData['ilo-ip'] = theServer['ilo-ip'] || '';
```
on this line and the ones that follow

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+      </div>
+    );
+  }
+
+  renderOsPasswordInput() {
+    return (
+      <InputLine
+        isRequired={this.state.isInstallOsChecked} inputName={'osInstallPassword'}
+        inputType={'password'} label={'server.pass.prompt'}
+        inputValue={this.state.osInstallPassword ? this.state.osInstallPassword : ''}
```
GarySmith: use `this.state.osInstallPassword || ''`

src/pages/ReplaceServer/PrepareReplace.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  updatePageStatus = (status) => {
+    this.setState({overallStatus: status});
+    if (status === STATUS.FAILED) {
+      this.setState({ invalidMsg: translate('server.replace.prepare.failure')});
+    }
+  }
+
+  goForward = (e) => {
```
GarySmith: Is this function needed? It should just call the parent's goForward without this, right?

src/pages/UpdateServers.js
```diff
@@ -0,0 +1,198 @@
...
+    this.state = {
+      model: this.props.model
+    };
+  }
+
+  componentWillReceiveProps(newProps) {
+    //deal with timing issue when load model
+    this.setState({
+      model : newProps.model,
+    } , () => {
```
GarySmith: As mentioned above, calling `this.setState` in the callback to `this.setState` is problematic and should be avoided.

GarySmith (changes requested):
This is great work, and I was amazed to see how functional it is already. Just need to address some stuff.

guglo:
@GarySmith @jeremy-moffitt  thanks for the review, I will go through the comments and address them when I get back to the further development.

src/styles/pages/ServerRoleSummary.less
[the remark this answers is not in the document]
guglo: saw you fixed it thanks.

src/InstallWizard.js
guglo: those are the sets for global and persisted variables for install and update. Update need some extra vars.

src/InstallWizard.js
guglo: I will use a const to indicate if it is update to make it a little clearer

src/InstallWizard.js
guglo: yeah...tried to reuse the code in InstallWizard. I will add more comments.

src/components/CollapsibleTable.js
guglo: see your point. maybe next round to cleanup.

src/components/PlaybookProcess.js
guglo: will update

src/pages/CloudDeployProgress.js
guglo: no it is not used in PlaybookProgress

src/pages/ServerRoleSummary.js
guglo: this is just clean up.

src/styles/components/contextmenu.less
guglo: will make it a little bigger

src/utils/DynamicPagesFinder.js
guglo: will update

src/InstallWizard.js
```diff
@@ -63,31 +62,83 @@ class InstallWizard extends Component {
       model: Map(),           // immutable model
       connectionInfo: undefined, // config info for external discovery services
       deployConfig: undefined, // cloud deployment configuration
+
+      //deal with update
+      currentMenuName: this.props.menuName,
+      currentOperation: undefined,
+      pages: undefined,
```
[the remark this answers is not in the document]
guglo: I will add comments to explain

src/InstallWizard.js
```diff
@@ -63,31 +62,83 @@ class InstallWizard extends Component {
       model: Map(),           // immutable model
       connectionInfo: undefined, // config info for external discovery services
       deployConfig: undefined, // cloud deployment configuration
+
+      //deal with update
+      currentMenuName: this.props.menuName,
```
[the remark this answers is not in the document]
guglo: yeah...i will add a const to indicate if it is update. The goal is to reuse the code in InstallWaizard.

src/InstallWizard.js
```diff
@@ -126,6 +172,20 @@ class InstallWizard extends Component {
       });
   }
 
+  componentWillMount = () => {
```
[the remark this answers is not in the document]
guglo: componentWillMount is for update and componentDidMount is for install. loading in componentDidMount for update has some issues. so leave the install load in componentDidMount as it is.  I will do some more testing to see if I can move update into componentDidMount

src/InstallWizard.js
```diff
@@ -63,31 +62,83 @@ class InstallWizard extends Component {
...
-    const forcedReset = window.location.search.indexOf('reset=true') !== -1;
+    const forcedReset =
+      this.props.menuName ? false : window.location.search.indexOf('reset=true') !== -1;
 
     // Load the current state information from the backend
-
     // To simulate a delay in startup and to be able to see the initial loading mask,
-    //    uncomment the following line and the line with the closing parenthesis two lines later
-    // new Promise(resolve => setTimeout(resolve, 1000)).then(() =>
+    // uncomment the following line and the line with the closing parenthesis two lines later
```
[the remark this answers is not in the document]
guglo: will update

src/InstallWizard.js
```diff
@@ -126,6 +172,20 @@ class InstallWizard extends Component {
       });
   }
 
+  componentWillMount = () => {
+    //update
+    if(this.props.menuName) {
+      this.loadModelandProgress(this.loadUpdateProgress);
```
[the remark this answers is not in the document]
guglo: will update

src/InstallWizard.js
```diff
@@ -274,6 +352,50 @@ class InstallWizard extends Component {
     }
   }
 
+  //this is triggered when start the update processes
+  //for example click repalce button on Replace server modal
+  startUpdate = (operation, pages, theProps) => {
+    //build steps dynamically
+    let steps = pages.map(page => { return {'name': page.name}; });
+    //get the first one start
+    steps[0].stepProgress = 2; //in progress
```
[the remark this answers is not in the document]
guglo: will udpate

src/components/CollapsibleTable.js
```diff
@@ -97,7 +165,7 @@ class CollapsibleTable extends Component {
       .map(g => new Map({
         'groupName': g,
         'members': groupMap.get(g),
-        'isExpanded': this.props.expandedGroup.includes(g)}))
+        'isExpanded': this.props.expandedGroup ? this.props.expandedGroup.includes(g) : false}))
```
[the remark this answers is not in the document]
guglo: will update

src/components/CollapsibleTable.js
```diff
@@ -202,14 +292,41 @@ class CollapsibleTable extends Component {
     );
   }
 
+  renderReplaceServerModal() {
+    let title =
+      translate('server.replace.heading') + (this.state.activeRowData ? '- ' +
```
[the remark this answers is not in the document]
guglo: will update

src/components/ContextMenu.js
```diff
@@ -58,16 +23,14 @@ class ContextMenu extends Component {
       show: this.props.show,
       location: this.props.location
     };
-
-    this.handleClickOutside = this.handleClickOutside.bind(this);
   }
 
   componentDidMount() {
-    document.addEventListener('click', this.handleClickOutside);
+    document.addEventListener('click', this.handleCloseMenu);
```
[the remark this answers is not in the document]
guglo: this handles click outside menu area. when user clicks menu outside area, the menu closes. will update comments

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+
+    this.allInputsStatus = {
+      'ilo-user': INPUT_STATUS.UNKNOWN,
+      'ilo-password': INPUT_STATUS.UNKNOWN,
+      'ilo-ip': INPUT_STATUS.UNKNOWN,
+      'mac-addr': INPUT_STATUS.UNKNOWN
+    };
+
+    this.state = {
+      isFormValid: false,
```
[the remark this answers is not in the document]
guglo: yeah...will update

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+    this.updateFormValidity(props, isValid);
+    let value = e.target.value;
+    if (isValid) {
+      if(props.inputName !== 'osInstallPassword') {
+        let newData = this.state.replaceData.set(props.inputName, value);
+        this.setState({replaceData: newData});
+      }
+      else {
+        this.setState({osInstallPassword: value}, () => {
+          this.setState({isFormValid: this.isFormTextInputValid()});
```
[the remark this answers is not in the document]
guglo: will update

src/components/ReplaceServerDetails.js
```diff
@@ -0,0 +1,321 @@
...
+    if(this.state.isInstallOsChecked) {
+      isAllValid = isAllValid && this.state.osInstallPassword && this.state.osInstallPassword !== '';
+    }
+
+    return isAllValid;
+  }
+
+  updateFormValidity = (props, isValid) => {
+    this.allInputsStatus[props.inputName] = isValid ? INPUT_STATUS.VALID : INPUT_STATUS.INVALID;
+    this.setState({isFormValid: this.isFormTextInputValid()});
```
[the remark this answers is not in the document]
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update to simplify and clean up

src/components/ReplaceServerDetails.js
guglo: k

src/components/ReplaceServerDetails.js
guglo: k

src/components/ReplaceServerDetails.js
guglo: will see what I can do

src/components/ReplaceServerDetails.js
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update

src/pages/ReplaceServer/PrepareReplace.js
guglo: will update

src/pages/UpdateServers.js
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update

src/InstallWizard.js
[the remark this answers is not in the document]
jeremy-moffitt: was there any luck with this?

jeremy-moffitt (approved):

src/InstallWizard.js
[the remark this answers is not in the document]
jeremy-moffitt: I missed the "if" statement with the IS_UPDATE conditional... I am curious about why it works differently, but at least its not loading the model twice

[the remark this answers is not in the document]
guglo: Tried a couple of times today...wired, the problem seems go away...will keep the loading in the original componentDidMount.

GarySmith:
The approach of folding in the update functionality into the existing InstallWizard seems problematic; the InstallWizard is built around the assumption that only one user can be running the application at a time. In day2 scenarios, it is possible that more than one user is logged in at the same time, so this is going to be problematic if this wizard is used at the same time by different users (for example two unrelated. If we are going to merge this in its current state, then we should add a new story to the backlog to refactor this.

src/components/ModelServerDetails.js
```diff
@@ -13,7 +13,8 @@
 * limitations under the License.
 **/
 import React, { Component } from 'react';
-import { translate } from '../../localization/localize.js';
+import { translate } from '../localization/localize.js';
+import { maskPassword } from '../utils/ModelUtils.js';
```
GarySmith: It does not make sense to import this `maskPassword` function and then overwrite it on line 33.

src/pages/BaseWizardPage.js
```diff
@@ -73,7 +74,17 @@ class BaseWizardPage extends Component {
     return false;
   }
 
-  renderNavButtons() {
+  closeUpdateProcess(e) {
```
GarySmith: These functions that are specific to the update do not belong in this class, which is a base class for any page that appear in the wizard.

jeremy-moffitt:
I dont think the ardana framework supports parallel updates... it does pose an interesting question for the multi-user scenario, I guess we'd have to lock change functionality while a change is going on if we're not assuming that any user logged in during an update is party to the update

src/utils/InputValidators.js
```diff
@@ -63,6 +79,21 @@ export function MacAddressValidator(macAddress) {
...
 
+  if(props && props.exist_mac_addresses && props.exist_mac_addresses.includes(macAddress)) {
+    retValue.isValid = false;
+    retValue.errorMsg = translate('input.validator.macaddress.exist.error');
+  }
+
+  if(props && props.exist_availservers_mac_addr_objs) {
+    let find =
+      props.exist_availservers_mac_addr_objs.find(mac => mac['mac-addr'] === macAddress);
+    if (find) {
```
GarySmith: This logic is broken. `find()` returns the array element, so if the first element (`[0]`) were a match, this logic would not detect it. You should probably replace `find()` with `some()`, which returns `true` if there is a match (and also rename the `find` variable to something like `found`)

src/utils/constants.js
```diff
@@ -36,6 +36,10 @@ export const MODEL_SERVER_PROPS =  [
   'id', 'ip-addr', 'mac-addr' ,'server-group' ,'nic-mapping' ,'ilo-ip' ,'ilo-user', 'ilo-password'
 ];
 
+export const REPlACE_SERVER_PROPS =  [
```
GarySmith: Typo here (the l in REPlACE should be in uppercase)

GarySmith:
@jeremy-moffitt I would not necessarily rule out concurrent updates (since concurrent playbook runs are certainly allowed). But regardless of that, the install wizard was designed so that if a second user happens to connect (of the single user re-connects), then they are both shown the same screen. I don't think that we would want that behavior in a multi-user UI.

src/components/ModelServerDetails.js
guglo: ah...function at line 33 was moved the ModelUtils, supposed to be removed. Will remove it here.

src/pages/BaseWizardPage.js
guglo: All the update pages reused most of the components from the BaseWizardPage,  Do you mean I need to have a another say BaseUpdateWizardPage which extends from BaseWizardPage to have closeUpdateProcess and cancelUpdateProcess moved there?

src/utils/DynamicPagesFinder.js
```diff
@@ -0,0 +1,49 @@
...
+ * specified. <menuName_operationName> is the key to find all the possible pages
+ * defined for the menu. steps will decide which pages we need for this update
+ * process.
+ *
+ * @param menuName           the menu name , for example servers
+ * @param operationName      the operation name, for example ReplaceServer
+ * @param steps              the steps for the process.
+ * @returns pages            assembled page components based on steps
+ */
+export function getPages(menuName, operationName, steps) {
```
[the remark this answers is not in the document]
guglo: getPages is called from InstallWizard During loadUpdateProgress.  It uses menuName, operation and steps to determine what pages are needed to load. Even UpdateServerPages are specific for the update servers, need a common place where wizard can pick up what is the left.

src/utils/constants.js
guglo: will update

guglo:
@GarySmith @jeremy-moffitt  regarding multi users ...assume we don't allow concurrent updates. However at this point, no restrictions. Need to have a restriction when update is in progress.

guglo:
I agree that we need have a design for how to address multiple administrators access and perform operations in  the day2UI. I think it might be better to address multiple users in the separate commit and refactor the code if we need to. At this point, I will record the current user name once user kicks off the update process so it could be used to restrict pages for readonly purpose for other users later.

src/utils/InputValidators.js
guglo: my element is an array of objects.  I can not use some() since I want the object back so I can use the serverId to construct my message. I think the better way is to change if(find) to be if (found !== undefined)

GarySmith: You are right, it was my misunderstanding. The code here is fine.

src/pages/BaseWizardPage.js
GarySmith: Yes, exactly. The base class should only contain things that are common to all pages

GarySmith (approved):

GarySmith:
This appear to address the concerns I had and to implement the improved design that Gloria and I discussed last week. Since github cannot gracefully handle forced pushes in reviews, it is nearly impossible to tell exactly what changed between this version and the previous one.

PR #116, opened 2018-07-27
SCRD-3953 Update build scripts

GarySmith:
Add script for building tarball of node_modules dependencies.  Remove
obsolete build_deps.sh and change run_dist.sh to call build_ui.sh
directly.
Fully leverage webpack to build the dist folder with all of the
build artifacts.
Remove index.html and generate it programatically from webpack.
Remove material icon fonts, bootstrap css, and jquery from source and
use ones obtained from npm.
Leverage webpack to automatically watch and build less files.
Support and detect development mode vs. production mode in webpack
config.
Convert config.json files to javascript and load statically rather than
via a REST call at run-time.
Generate source maps for css so to facilitate inspection in browser
development tools.


GarySmith:
1. This is caused by having an old nodeJS binary. The version used by the IBS build, 6.14.3, works correctly.
2. I don't think this problem is related to this change. I also saw something similar to this once (before making this change) but could not reproduce it either.

jeremy-moffitt:
It might make sense to add the following to the .gitignore file:
node_modules.tar.bz2
yarn.lock

otherwise looks great!

PR #120, opened 2018-08-08
Add Service Information page

jgravel:
SCRD-3484 Added 'Service Information' page which is the page under the 'Information' tab off of the Services menu. The page contains a table with four columns: Name, Description, Endpoints, and Regions. There's supposed to be the fifth column which is the 'Status' column, but we need the status data from Monasca which is not available at the moment. Once the Monasca data is available, we will need to add the Status column to the table.

Also updated how the navigation menu works by getting rid of the landing page and setting the first tab of each menu to be the default page when the menu was selected.

### DIFF
--- a/src/components/NavMenu.js
+++ b/src/components/NavMenu.js
@@ -30,7 +30,7 @@ class NavMenu extends Component {
         path={e.slug}
         children={({ match }) => (
           <li >
-            <Link className={match ? 'active' : ''} to={e.slug}>{e.name}</Link>
+            <Link className={match ? 'active' : ''} to={e.items[0].slug}>{e.name}</Link>
           </li>)}
       />
     ));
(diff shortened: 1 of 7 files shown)


src/components/NavMenu.js
```diff
@@ -30,7 +30,7 @@ class NavMenu extends Component {
         path={e.slug}
         children={({ match }) => (
           <li >
-            <Link className={match ? 'active' : ''} to={e.slug}>{e.name}</Link>
+            <Link className={match ? 'active' : ''} to={e.items[0].slug}>{e.name}</Link>
```
jeremy-moffitt: the change to navigation matches what we discussed, it might have made sense to do 2 separate commits for tracking purposes, but probably not a big deal

jgravel: I'd be happy to split the change in NavMenu.js out but that will be a one-liner and the behavior will not be quite complete. When you first login, the first tab will not be underlined (because it needs the change in login.js). Clicking around on other menus after that will give you the right behavior. This will be temporary until the rest of this change merges anyhow, so it shouldn't be too much of an issue. If you're ok with that, I'll make the change.

PR #121, opened 2018-08-10
Fixed discover tab margin issue

ghost:
The discovery/manual table has a wide margin in the installer, fix will
prevent the conflict styling class


jeremy-moffitt (approved):
thanks for finding/fixing this

PR #122, opened 2018-08-14
Add alarm summary graph component

GarySmith:
Add graph component and corresponding test page


GarySmith:
I'll change it to 12. 
As with the speedometer demo page, the border is not part of the component, but was added in order to clearly identify the size of the component. I would expect that the page that includes this and any other components would set the appropriate margin/padding to avoid them abutting to close to their edges. This border box is very useful to have in the demo page, and was useful in diagnosing size and layout issues during development. Do you have any suggestions for how I can keep this feature in the demo page without being confusing?

PR #123, opened 2018-08-14
Mention new build_deps.sh in readme

GarySmith:


GarySmith:
Oh, good catch. I didn't read anything other than the last paragraph

PR #124, opened 2018-08-14
Fixed issue when installation progress.json is present

ghost:
If progress.json left from installation, the steps could mess up the servers page loading in day2UI, add a restriction for day2UI progress loading.

### DIFF
--- a/src/InstallWizard.js
+++ b/src/InstallWizard.js
@@ -146,7 +146,7 @@ class InstallWizard extends Component {
   loadUpdateProgress = (responseData) => {
     //TODO if refresh with default url or logout due to timeout
     //need to find a way to set the menu
-    if(responseData && responseData.steps) {
+    if(responseData && responseData.currentMenuName && responseData.steps) {
       let steps = responseData.steps;
       let pages =
         this.getPages(steps);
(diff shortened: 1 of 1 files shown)


jeremy-moffitt (approved):
LGTM, tried it out and seemed to work ok... though I'm not 100% clear what the change is... only interpret the progress if currentMenuName is set... it definitely works, I just dont understand how

guglo:
@jeremy-moffitt when loadUpdateProgress, if there is no currentMenuName to present...then it treats as there are no progresses and no steps so it will ignore the steps left from installation.  when user starts replace process in Day2UI, it records the currentMenuName into the progress  and record its own progress stuffs (steps, playbooks...).

GarySmith (commented):
While this fix seems to work, it deserves some kind of explanation in the code about what this odd check is for.

guglo:
@GarySmith, ok since Jeremy also asks the question, I will add some comments at the code.

guglo:
@GarySmith  Are you happy with the comments update? Since can not rocketchat guess ping you here :)

GarySmith (approved):
Thanks for adding the explanation. Looks great.

PR #130, opened 2018-08-18
Fixed the error message close issue

ghost:
This checkin fixed the issuse when messages in the assign server role
page or in the connection info modal can not be closed by click x.
(SCRD-4453)

To test:
1. import csv file with item missing ip address to make the error message show up.
2. do discovery, add a connetion and click test button to have a sucessful message


jeremy-moffitt (approved):
works! I'm confused why return {messages: prevState.messages.slice().splice(ind,1)}  doesnt work, but I tried it and I'm guessing that the messages are immutable or something due to being part of the state... still, the slice should return a shallow copy which is then spliced... /ohwell

PR #131, opened 2018-08-20
Add control plane topology page

GarySmith:
Add the control plane page into the topology section of the menu.  This
page is very close to the control plane page that is emitted by the
configuration processor.

### DIFF
--- a/src/Deployer.js
+++ b/src/Deployer.js
@@ -54,7 +54,7 @@ class Deployer extends Component {
     if (this.state.isSecured === undefined) {
       // If the REST call has not yet completed, then show a loading page (briefly)
       defaultPath = (
-        <div className="loading-message">{translate('wizard.loading.pleasewait')}</div>
+        <div className="loading-message">{translate('loading.pleasewait')}</div>
       );
     } else if (this.state.isSecured) {
       if (! getAuthToken()) {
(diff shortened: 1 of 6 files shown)


GarySmith:
Note that some of the fields are created as html links in order to mimic the config processor page, but these links are not (yet) valid since the other topology pages have not been created.  These links will be updated when those pages are created.

GarySmith:
I hadn't planned to put boxes since we are not using them anywhere in the app so far. These could be added.
The long load is due to the fact that the file that is loaded on the back end is pretty large (the midsize model's is more then 6MB). As an performance improvement I plan to be able to load it once and share it between the various topology tabs to avoid the delays.

jeremy-moffitt:
Hmm ok, I'll have to look at it again, having seen the previous version I thought that the groupings from the boxes made it more obvious which things were associated with each other... on the performance thing, I wonder if we could compress the file before sending it across the wire

GarySmith:
The compression *may* help a bit, but I think the bulk of the time is actually spent loading the file, converting the file into python, than converting it into json to be sent. As a test, I have a little program that does those steps without actually transferring the file, and it takes about 7.9 seconds to do just that. The UI actually only waits 8.1 seconds, so I think the transfer time is only about 0.2 seconds.

Agreed on the boxes and grouping. At this point I was more concerned with being able to reproduce all of the data and leaving the details of formatting for later. I guess I should have been more clear on that. I think we still probably need to talk through this formatting and decide what is best before reproducing those odd boxes and and shading in the existing html.

PR #132, opened 2018-08-21
Added initial adding compute node landing page

ghost:
This checkin includes:

1. Added the landing page for adding compute nodes by reusing the
assign server role page

2. Updated the various components in assign server role page  with
the followings:
* only show the compute role in the accordion
* Discovery should still work as it is in the installtion
* Manual add servers modal only shows none and compute roles
* Import servers csv only imports servers without roles or with compute roles
* if the servers imported with the same id or ip-addr, ilo-ip, or mac-addr as
  the servers in the model,  they will be ignored.
* Still need a way to differentiate deployed servers in the model and
newly added servers, temporarily make the edit and delete actionunavaialble
in the role accordion.


src/InstallWizard.js
```diff
@@ -152,12 +152,18 @@ class InstallWizard extends Component {
...
-      updateStates.pages = pages;
+      if(responseData.steps) {
+        let steps = responseData.steps;
+        let pages =
+          this.getPages(steps);
+        updateStates.pages = pages;
+      }
+      else {
+        // if there is no progress, will record the menuName passed in
+        updateStates.currentMenuName = this.props.menuName;
```
jeremy-moffitt: this if block confuses me a little... its checking for steps, but sets currentMenuName... I assume thats intentional, but its a bit confusing

src/localization/bundles/en.json
```diff
@@ -259,6 +260,7 @@
     "server.import.update.error": "Failed to save updated imported server {0}",
     "server.import.add.error": "Failed to save newly imported servers",
     "server.import.server.confirm": "The imported servers will overwrite previously imported servers if they have the same ID.\n If imported servers have the same ID as the manually added servers or discovered servers, servers with duplicate ID could show up. Servers that have been assigned a role and have duplicate ID will be highlighted in red. You can edit the server to update the ID.\n\n Do you still want to import servers? ",
+    "server.import.server.confirm.limit.conflict": "The imported servers withtout role will overwrite previously imported servers without role if they have the same ID.\n If imported servers have the same ID as the manually added servers or discovered servers, servers with duplicate ID could show up. Servers with roles that don't belong to the target roles will be ignored. Servers that have been assigned a role and have a duplicate ID, or IP address or MAC address or IPMI address will be ignored. \n\n Do you still want to import servers? ",
```
jeremy-moffitt: s/withtout/without/

src/pages/AssignServerRoles.js
```diff
@@ -1266,17 +1304,28 @@ class AssignServerRoles extends BaseWizardPage {
...
         ondragEnterFunct={this.highlightDrop}
         ondragLeaveFunct={this.unHighlightDrop}
         allowDropFunct={this.allowDrop}
-        serverRoles={getServerRoles(this.props.model)}
-        tableId='rightTableId' checkDupIds={dupIds}
-        editAction={this.handleShowEditServer}
-        viewAction={this.handleShowServerDetails}
-        deleteAction={this.handleDeleteServer}>
+        serverRoles={getServerRoles(this.props.model, this.props.rolesLimit)}
+        tableId='rightTableId' checkDupIds={dupIds} {...theProps}
```
jeremy-moffitt: I'd probably move {...theProps} to its own line so its more obvious that its a standalone set of key/value pairs... maybe rename it as well... extraProps or something... "theProps" had me confused whether it was duplicating the current objects props or something passed down to the next object

jeremy-moffitt (changes requested):
overall looks fairly clean, couple nitpicks to address

src/InstallWizard.js
guglo: @jeremy-moffitt  will try to update the comments

src/pages/AssignServerRoles.js
guglo: ok will update

src/localization/bundles/en.json
guglo: ok will update

src/InstallWizard.js
guglo: ok, not sure update will talk about it first.

jeremy-moffitt (approved):
looks fine to me

GarySmith (commented):
The Install OS and Deploy buttons are available but don't seem to do anything. Is this expected?

guglo:
@GarySmith  this is incremental checkin to avoid a large amount of code commit like last time. this is just initial landing page checkin.

GarySmith:
So I'll understand that answer to be "yes, this is expected"

src/components/ServerUtils.js
```diff
@@ -113,6 +113,7 @@ export class ServerRolesAccordion extends Component {
       let isOpen = (idx === this.state.accordionPosition);
       let valid = isRoleAssignmentValid(role);
       let triggerClass = valid ? '' : 'has-error';
+      let innerClass = this.props.mode && this.props.mode  === 'addserver' ? 'addserver': '';
```
GarySmith: The initial check before the && is unnecesary. It would be more straightforward to use something like:
`   let innerClass = this.props.mode || '';
`

src/pages/AssignServerRoles.js
```diff
@@ -336,6 +351,22 @@ class AssignServerRoles extends BaseWizardPage {
...
     }
 
+    // if it is update and imported server has the same id or ip-addr or mac-addr or ilo-ip
+    // as any server in the model already, it will be ignored
+    if(this.props.mode && this.props.mode === 'addserver') {
+      servers = servers.filter(server => {
+        const found = model.getIn(['inputModel', 'servers']).some(svr => {
+          return (
+            svr.get('id') === server['id'] ||
+            (svr.get('ip-addr') && svr.get('ip-addr') === server['ip-addr']) ||
```
GarySmith: These checks before the && are also unnecessary. It should just be `svr.get('ip-addr') === server['ip-addr']` and so on

src/pages/AssignServerRoles.js
```diff
@@ -1373,50 +1424,67 @@ class AssignServerRoles extends BaseWizardPage {
...
+        {this.props.mode === undefined && this.renderInstallContentHeading()}
+        <YesNoModal show={this.state.showDeleteServerConfirmModal} title={translate('warning')}
+          yesAction={this.deleteServer}
+          noAction={() => this.setState({showDeleteServerConfirmModal: false})}>
+          {translate('server.delete.server.confirm', serverId)}
+        </YesNoModal>
+        <YesNoModal show={this.state.showImportServerConfirmModal} title={translate('warning')}
+          yesAction={this.saveImportedServers}
+          noAction={() => this.setState({showImportServerConfirmModal: false, importedResults: {}})}>
+          {this.props.mode && this.props.mode === 'addserver' ?
```
GarySmith: again, unnecessary check before the &&

src/pages/AssignServerRoles.js
```diff
@@ -1373,50 +1424,67 @@ class AssignServerRoles extends BaseWizardPage {
...
+        </div>
+      </div>
+    );
+  }
+
+  //gloria
   render() {
     let serverId = (this.state.activeRowData && this.state.activeRowData.id) ? this.state.activeRowData.id : '';
+    const contentClass =
+      'wizard-content' + (this.props.mode && this.props.mode === 'addserver' ? ' smaller-margin' : '');
```
GarySmith: unnecessary check before the &&

src/pages/AssignServerRoles.js
```diff
@@ -427,6 +458,14 @@ class AssignServerRoles extends BaseWizardPage {
         server['uid'] = genUID('import');
       }
 
+      if(this.props.mode && this.props.mode === 'addserver') {
```
GarySmith: unnecessary check before &&

src/utils/ModelUtils.js
```diff
@@ -272,3 +285,13 @@ export function maskPassword(pass) {
 
   return '*'.repeat(pass.length);
 }
+
+// check theRole for example COMPUTE-ROLE matches rolesLimit for example
+// COMPUTE
+export function matchRolesLimit(theRole, rolesLimit) {
+  let match = rolesLimit.some(
+    role =>
+      theRole.includes(role.toUpperCase()) || theRole.includes(role.toLowerCase())
```
GarySmith: What is going on here with the case of these roles? IIRC, these are not being entered by the user, so they should never differ in case. If turns out that roles are some special thing that can have a variety of cases and still be considered equal, then we should have some utility method for comparing them and use that everywhere in our code base.

GarySmith (changes requested):

src/components/ServerUtils.js
guglo: @GarySmith  I was not sure if the mode will be eventually called addserver, just want to be flexible if there other mode.

src/pages/AssignServerRoles.js
guglo: same response as above...mode is not necessary 'addserver'

src/pages/AssignServerRoles.js
guglo: ip-addr might be ok to remove since import csv logic already make sure it is not empty. However for mac-addr, ilo-ip, could be empty to start with if user discovered something,  if both of them empty, it could be treated as equal.  
svr.get('ilo-ip') is undefined server['ilo-ip'] is undefined...svr.get('ilo-ip') === server['ilo-ip'] is true

src/pages/AssignServerRoles.js
guglo: will update

src/utils/ModelUtils.js
guglo: Hmmm, my understanding is role could be defined by user...so case could be various

src/pages/AssignServerRoles.js
guglo: k, will update

src/pages/AssignServerRoles.js
guglo: ok will update

src/pages/AssignServerRoles.js
guglo: will update comments

GarySmith (approved):

PR #133, opened 2018-08-21
Add Run Status Playbook action for services

jgravel:
SCRD-3485 Added a drop down menu that contained 'Details' and 'Run Status Playbook' actions for each service in the Service Information page. Then implemented the modal dialog that showed the streaming of result from running status playbook for the selected service. The Details menu doesn't do anything at the moment.

### DIFF
--- a/src/components/Buttons.js
+++ b/src/components/Buttons.js
@@ -40,7 +40,7 @@ function CloseButton(props) {
   return (
     <ActionButton
       clickAction={props.clickAction}
-      displayLabel={props.displayLabel || translate('common.close')}
+      displayLabel={props.displayLabel || translate('close')}
       isDisabled={props.isDisabled}
     />
   );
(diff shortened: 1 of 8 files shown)


jgravel:
@jeremy-moffitt  How did you run with site.yml? I'd like to be able to reproduce the behavior. 

Regarding the resizability of the modal, the current modal component we're using in our modals are from react-bootstrap which does not support such capability. There are two alternatives I found from a quick google search: one is through jQuery and the other one is through a library call react-modal-resizable-draggable. I haven't looked closely enough into either one of them but I'm going to try jQuery first. If that doesn't work, I'll try the other one.

jeremy-moffitt:
I changed:
`const playbookName = this.state.selectedService + '-status'`
to
`const playbookName = 'site'

that made it run the site.yml playbook (which the dev ardana-service has a mock run for)... since I couldnt run a real -status.yml playbook locally

PR #135, opened 2018-08-22
Add region topology page

GarySmith:
Add region page in the topology section.  Introduce caching so that
toggling between topology tabs does not cause a re-load.  Use
componentDidMount instead of componentWillMount for loading data
per the guidance in the react documents:
https://reactjs.org/docs/react-component.html#componentdidmount .
Enable linking to anchors on different pages.

### DIFF
--- a/src/pages/topology/ControlPlanes.js
+++ b/src/pages/topology/ControlPlanes.js
@@ -16,6 +16,8 @@ import React, { Component } from 'react';
 import { translate } from '../../localization/localize.js';
 import '../../styles/deployer.less';
 import { getInternalModel } from './TopologyUtils.js';
+import { ErrorBanner } from '../../components/Messages';
+import { LoadingMask } from '../../components/LoadingMask';
 
 const Fragment = React.Fragment;
 /*
@@ -28,28 +30,11 @@ class ControlPlanes extends Component {
     super(props);
 
     this.state = {
-      model: undefined
+      model: undefined,
+      errorMessage: undefined,
     };
 
-    this.cloud_internal = undefined;
-    this.control_planes = undefined;
-    this.servers = undefined;
     this.server_by_hostname = {};
-  }
-
-  init = () => {
-    if (this.state.model) {
-      this.cloud_internal = this.state.model['internal'];
-      this.cp_topology = this.cloud_internal['cp-topology'];
-      this.control_planes = this.cp_topology['control_planes'];
-      this.servers = this.cloud_internal['servers'];
-
-      for (const s of this.servers) {
-        if (s['hostname']) {
-          this.server_by_hostname[s['hostname']] = s;
-        }
-      }
-    }
   }
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


jeremy-moffitt:
note for other reviewers, you'll need to do an `npm install`  to get react-router-hash-link  , remind me after we merge this to update the node_modules tarball again

src/pages/topology/Regions.js
```diff
@@ -0,0 +1,139 @@
...
+            // highlight it in the displayed list
+            let found = false;
+            for(const component of region_svcs[svc]) {
+              if (advertised[name].includes(component)) {
+                found = true;
+                break;
+              }
+            }
+            if (found) {
+              return <div key={svc}><b>{svc}</b></div>;
```
guglo: the found one in bold with other one in slightly lighter found looks a little odd at the first sight. Might need UX input for that to make it obvious to user that one is advertised through keystone  and one is not.

src/pages/topology/ControlPlanes.js
```diff
@@ -53,19 +52,18 @@ class ControlPlanes extends Component {
...
-        this.setState({
-          model: yml});
+        // Force a re-render if the page is still shown (user may navigate away while waiting)
+        if (this.refs.control_planes)
+          this.setState({
+            model: yml});
       })
       .catch((error) => {
-        // console.log(error);
+        console.log(error); // eslint-disable-line no-console
```
guglo: It would be better to handle it instead of console log it. Because if there is an error here...it  stucks at '
Loading... Please wait' ...have to open console to see the error. Add an error banner or error message

GarySmith: Good point. Will update with a fix

src/pages/topology/Regions.js
GarySmith: I agree that this way of using bold to represent items advertised by keystone is not great -- I was simply reproducing here what was done in the original pages produced by the config processor. It, as well as all of these topology pages, needs some UX help.

guglo: agree,  thanks.

guglo:
@GarySmith  downloaded the latest change, the region tab shows empty example now

src/utils/RouteConfig.js
```diff
@@ -69,11 +70,11 @@ export const routes = [
   { name: translate('topology'), slug: '/topology',
     items: [
       { name: translate('control_planes'), slug: '/topology/control_planes', component: ControlPlanes },
-      { name: translate('regions'), slug: '/topology/regions', component: Regions },
-      { name: translate('services'), slug: '/topology/services', component: Example },
-      { name: translate('networks'), slug: '/topology/networks', component: Example },
-      { name: translate('servers'), slug: '/topology/servers', component: Example },
-      { name: translate('server_groups'), slug: '/topology/server-groups', component: Example },
+      { name: translate('regions'), slug: '/topology/regions', component: Example },
```
jeremy-moffitt: git merged this the wrong way for what you want, that explains why the example content is showing up instead of the region topology

src/pages/topology/Regions.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  render () {
+    return (
+      <div ref="regions" className='wizard-page'>
+        <LoadingMask show={!this.state.model && !this.state.errorMessage}/>
+        <div className='wizard-content'>
+          {this.state.model && this.render_regions()}
+        </div>
+        <div className='notification-message-container'>
```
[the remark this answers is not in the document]
GarySmith: That's good to know. I didn't realize it was supposed to look different

PR #137, opened 2018-08-23
Fixed close issue for error msg of login

ghost:
This is the continue fixing  of SCRD-4453

### DIFF
--- a/src/styles/components/messages.less
+++ b/src/styles/components/messages.less
@@ -23,6 +23,9 @@
   position: absolute;
   top: 1em;
   right: 1em;
+  &.login {
+    z-index: 1111;
+  }
 }
 
 .loading-message {
--- a/src/styles/components/messages.less
+++ b/src/styles/components/messages.less
@@ -14,17 +14,15 @@
 **/
 @import "./../palette.less";
 
-.notification-message {
-  z-index: 1111;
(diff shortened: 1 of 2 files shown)


jeremy-moffitt (dismissed):
this works, and seems simple enough, its a little unfortunate that the z-index is set in 2 different related css classes

guglo:
this fixes it but it is not the best fix, will figure out the a better way.

guglo:
@jeremy-moffitt  thanks for the update, tested it, it works well.

src/styles/components/messages.less
```diff
@@ -14,17 +14,15 @@
...
-}
 
 .notification-message-container {
   position: absolute;
   top: 1em;
   right: 1em;
-  &.login {
+  &.login,
+  .notification-message{
     z-index: 1111;
```
GarySmith: Is there some rationale behind this very large z-index? I don't know much about z-indexes, but this looks like some special value.

jeremy-moffitt: its what was there before... not a great answer I know

jeremy-moffitt: I tried again with a z-index of 20, and it works for the login page... going to check the other places its used

PR #139, opened 2018-08-24
SCRD-4750 Add logout button to Day 2 UI

jeremy-moffitt:
SCRD-4750 Added the logout button to the masthead

### DIFF
--- a/src/components/NavMenu.js
+++ b/src/components/NavMenu.js
@@ -84,7 +84,7 @@ class NavMenu extends Component {
       redirectToLogin(false);
     }
 
-    const logout_tooltip = (<Tooltip id='logout' className='tooltip'>{translate('logout')}</Tooltip>);
+    const logout_tooltip = (<Tooltip id='logout'>{translate('logout')}</Tooltip>);
 
     return(
       <Router>


jgravel:
I agreed with Gary and I wished this change had been there when I worked on the login page. The change worked but I wonder if we should have a tool tip or some explanation for the logout button. The icon is self-explanatory for some people but having some text description would clear all the doubts. This is not a big issue. It can be done in a separate pull request. By the way, the SCC website uses the 'Logout' word instead of an icon.

jeremy-moffitt:
added a tooltip for the button, kept it as an icon for now... will include that on the items we ask for UX review in the near future

src/components/NavMenu.js
```diff
@@ -76,6 +79,13 @@ class NavMenu extends Component {
       });
     });
 
+    let logout = () => {
+      clearAuthToken();
+      redirectToLogin(false);
+    }
+
+    const logout_tooltip = (<Tooltip id='logout' className='tooltip'>{translate('logout')}</Tooltip>);
```
jgravel: I don't see the 'tooltip' class defined anywhere. Do we need it?

jeremy-moffitt: I copied that from some other code, I'll try without it

PR #140, opened 2018-08-24
Add services topology table

GarySmith:
Add service page in the topology section.
Add hyperlinks from services mentioned on the other topology pages to
  this page.
Refactor topology styles into a separate less file.


GarySmith:
@jgravel Thanks for noticing the lint errors.

PR #141, opened 2018-08-27
Add OpenStack Packages and SUSE Cloud Packages pages

jgravel:
SCRD-4309 Added two new pages under the Services menu: OpenStack Packages page and SUSE Cloud Packages page (for Ardana packages)

To test, you can point your shimurl in config.dev.js to Jack's machine at http://192.168.200.88:3000. I'm not sure what's the official name for Ardana is, so I'm using 'SUSE Cloud' for now. I can update if if/when we decide to use a different name.

### DIFF
--- a/src/pages/ArdanaPackages.js
+++ b/src/pages/ArdanaPackages.js
@@ -1,0 +1,78 @@
+// (c) Copyright 2018 SUSE LLC
+/**
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


src/pages/OpenStackPackages.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  render() {
+    let rows = [];
+    if (this.state.packages) {
+      rows = this.state.packages
+        .sort((a,b) => alphabetically(a.name, b.name))
+        .map((pkg, idx) => {
+          return (
+            <tr key={idx}>
```
GarySmith: It would be better to use `pkg.name` as the key rather than `idx`. See https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

src/pages/ArdanaPackages.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  render() {
+    let rows = [];
+    if (this.state.packages) {
+      rows = this.state.packages
+        .sort((a,b) => alphabetically(a.name, b.name))
+        .map((pkg, idx) => {
+          return (
+            <tr key={idx}>
```
GarySmith: It would be better to use something like `pkg.name` as the key rather than `idx`.

jgravel:
I updated the code per Gary's suggestion. Please take a look again.

PR #143, opened 2018-08-27
Updated to handle edit for adding a new compute server

ghost:
This is incremental checkin for SCRD-3476, it includes:

1. Updated Edit server modal to check the conflits for existing ip-addr,
mac-addr and ipmi-addr
2. Updated Edit server modal to hide the buttons for updating nic-mapping
and server-group for now
3. Udpated the assign server role page to have an option to set wipeDisk
4. Updated the server accordion to show in red for missing nic-mapping and
server-group in the servers
5. Updated the add server page to check invalid model and disable the deploy
button

Note:
1. install button and deploy button are still not functioning
2. Still can not differentiate what are the deployered servers and
what are newlly added servers.

### DIFF
--- a/src/components/EditServerDetails.js
+++ b/src/components/EditServerDetails.js
@@ -26,6 +26,8 @@ class EditServerDetails extends Component {
   constructor(props) {
     super(props);
 
+    this.IS_INSTALLL = this.props.mode ? false : true;
+
     this.allInputsStatus = {
       'id': INPUT_STATUS.UNKNOWN,
       'ip-addr': INPUT_STATUS.UNKNOWN,
@@ -197,7 +199,7 @@ class EditServerDetails extends Component {
           <div className='input-with-button'>
             <ListDropdown name={this.props.name} value={this.data[name]} moreClass={'has-button'}
               optionList={list} emptyOption={emptyOptProps} selectAction={handler}/>
-            {this.props.mode === undefined && this.renderButtonForDropDown(addAction, buttonLabel)}
+            {this.IS_INSTALLL && this.renderButtonForDropDown(addAction, buttonLabel)}
           </div>
         </div>
       </div>
@@ -253,8 +255,8 @@ class EditServerDetails extends Component {
       <div className='edit-server-details'>
         {this.renderServerContent()}
         {this.renderFooter()}
-        {this.props.mode === undefined && this.renderAddServerGroup()}
-        {this.props.mode === undefined && this.renderAddNicMapping()}
+        {this.IS_INSTALLL && this.renderAddServerGroup()}
+        {this.IS_INSTALLL && this.renderAddNicMapping()}
       </div>
     );
   }
--- a/src/components/EditServerDetails.js
+++ b/src/components/EditServerDetails.js
@@ -119,15 +119,27 @@ class EditServerDetails extends Component {
   }
 
   renderInput(name, type, isRequired, title, validate) {
-    let theProps = {};
+    let extraProps = {};
     if(name === 'id') {
-      theProps.ids = this.props.ids;
+      extraProps.ids = this.props.ids;
     }
+
+    if (name === 'mac-addr') {
+      extraProps['exist_mac_addresses'] = this.props.existMacAddressesModel;
+    }
+    if (name === 'ilo-ip') {
+      extraProps['exist_ip_addresses'] = this.props.existIPMIAddressesModel;
+    }
+    if (name === 'ip-addr') {
+      extraProps['exist_ip_addresses'] = this.props.existIPAddressesModel;
+    }
+
     return (
       <InputLine
-        isRequired={isRequired} inputName={name} inputType={type} label={title} {...theProps}
+        isRequired={isRequired} inputName={name} inputType={type} label={title}
         inputValidate={validate} inputValue={this.data[name] ? this.data[name] : ''} moreClass={'has-button'}
-        inputAction={this.handleInputChange} updateFormValidity={this.updateFormValidity}/>
+        inputAction={this.handleInputChange} updateFormValidity={this.updateFormValidity}
+        {...extraProps}/>
     );
   }
 
@@ -163,6 +175,13 @@ class EditServerDetails extends Component {
     );
   }
 
+  renderButtonForDropDown(addAction, buttonLabel) {
+    return (
+      <ActionButton type={'default'} clickAction={addAction} moreClass={'inline-button'}
+        displayLabel={translate(buttonLabel) + ' ...'}/>
+    );
+  }
+
   renderDropDown(name, list, handler, isRequired, title, buttonLabel, addAction) {
     let emptyOptProps = '';
     if(this.data[name] === '' || this.data[name] === undefined) {
@@ -178,8 +197,7 @@ class EditServerDetails extends Component {
           <div className='input-with-button'>
             <ListDropdown name={this.props.name} value={this.data[name]} moreClass={'has-button'}
               optionList={list} emptyOption={emptyOptProps} selectAction={handler}/>
-            <ActionButton type={'default'} clickAction={addAction} moreClass={'inline-button'}
-              displayLabel={translate(buttonLabel) + ' ...'}/>
+            {this.props.mode === undefined && this.renderButtonForDropDown(addAction, buttonLabel)}
           </div>
         </div>
       </div>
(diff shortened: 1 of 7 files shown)


src/components/EditServerDetails.js
```diff
@@ -163,6 +175,13 @@ class EditServerDetails extends Component {
     );
   }
 
+  renderButtonForDropDown(addAction, buttonLabel) {
+    return (
+      <ActionButton type={'default'} clickAction={addAction} moreClass={'inline-button'}
+        displayLabel={translate(buttonLabel) + ' ...'}/>
```
jeremy-moffitt: I'm confused whats going on here... I dont see any buttons on the page that have ... in the label, and I'm not sure that appending it here is the best route to go (maybe it is, I'm not sure... are there languages where the ... wouldnt be appropriate?)

src/components/EditServerDetails.js
```diff
@@ -178,8 +197,7 @@ class EditServerDetails extends Component {
           <div className='input-with-button'>
             <ListDropdown name={this.props.name} value={this.data[name]} moreClass={'has-button'}
               optionList={list} emptyOption={emptyOptProps} selectAction={handler}/>
-            <ActionButton type={'default'} clickAction={addAction} moreClass={'inline-button'}
-              displayLabel={translate(buttonLabel) + ' ...'}/>
+            {this.props.mode === undefined && this.renderButtonForDropDown(addAction, buttonLabel)}
```
jeremy-moffitt: maybe with this check for this.props.mode all over, it should be moved to its own function with a descriptive name... like isPreInstall or isInstalling or something... and just call that function everywhere that `this.props.mode === undefined`  is essentially checking if its going on during install or after

src/components/ReplaceServerDetails.js
```diff
@@ -202,26 +202,27 @@ class ReplaceServerDetails extends Component {
...
     }
     return (
       <InputLine
         isRequired={isRequired} inputName={name} inputType={type} label={title}
-        inputValidate={validate} {...theProps}
+        inputValidate={validate}
         inputValue={this.state.replaceData.get(name) || ''}
-        inputAction={this.handleInputChange} updateFormValidity={this.updateFormValidity}/>
+        inputAction={this.handleInputChange} updateFormValidity={this.updateFormValidity}
+        {...extraProps}/>
```
jeremy-moffitt: thank you for renaming theProps, I think this is clearer

src/components/EditServerDetails.js
guglo: @jeremy-moffitt  this is just an existing render function for those nic-mapping and server groups button in the edit dialog, I just moved out into a function so I can hide them since adding compute nodes along with adding nic-mapping or server group is not thought through.

src/components/EditServerDetails.js
guglo: this.props.mode means it is not in any addserver mode which day0 mode. yeah...it is all over the place , ok, will think about something  and update.

jeremy-moffitt: ah ok, I see, since its not a change, we can worry about the translation feedback some other time (it likely wont come up)

src/components/EditServerDetails.js
jeremy-moffitt: ah ok, I see, since its not a change, we can worry about the translation feedback some other time (it likely wont come up)

src/components/EditServerDetails.js
jeremy-moffitt: I just dont think its obvious to a casual observer what this check is doing... doing exactly the same thing inside another function that has a name would make it more obvious.... for example, in this case the code would read:
`{this.isInitialInstall() && this.renderButtonForDropDown(addAction, buttonLabel)}`  ... although hopefully someone can come up with a better function name than isInitialInstall  ;)  ... could reverse it and do `{!this.isPostInstall() && ...}`

guglo: ok will update call it  isInstall for now unless someone comes up with a better name

jeremy-moffitt (approved):
+1 , lgtm, thanks for changing the undefined checks, I think it reads a lot clearer! havent had a chance to test it with the latest changes

src/components/EditServerDetails.js
```diff
@@ -26,6 +26,8 @@ class EditServerDetails extends Component {
   constructor(props) {
     super(props);
 
+    this.IS_INSTALLL = this.props.mode ? false : true;
```
GarySmith: Can you fix this typo? INSTALL only has 2 L's

src/pages/AddServers.js
```diff
@@ -52,9 +56,16 @@ class AddServers extends BaseUpdateWizardPage {
     //TODO implement
   }
 
+  //check if we can deploy the new servers
   isDeployable = () => {
-    //TODO implement
-    return false;
+    if(this.state.model && this.state.model.size > 0) {
```
GarySmith: since `this.state.model` is set in the constructor, it seems like the check on the left of the && is unnecessary.

src/pages/AssignServerRoles.js
```diff
@@ -52,6 +53,9 @@ class AssignServerRoles extends BaseWizardPage {
   constructor(props) {
     super(props);
 
+    this.IS_INSTALLL = this.props.mode ? false : true;
```
GarySmith: typo here, too (should be IS_INSTALL). Actually this check is confusing because I thought that `this.props.mode` is some kind of string that contains either `addserver` or something else. If it is really a boolean, then can we rename it to something more meaningful like `this.props.addMode` or even `this.props.installMode`? That would also remove the need for this extra `IS_INSTALL` variable since the name would be more meaningful.

GarySmith (changes requested):

src/components/EditServerDetails.js
guglo: ok will fix

src/pages/AssignServerRoles.js
jeremy-moffitt: you could explicitly check for 'addserver' here, or check explicitly against undefined.... the logic makes sense, but I can see where at a glance it makes this.props.mode look like a boolean

guglo: thanks for clarify, will update

src/pages/AssignServerRoles.js
```diff
@@ -53,7 +53,7 @@ class AssignServerRoles extends BaseWizardPage {
   constructor(props) {
     super(props);
 
-    this.IS_INSTALLL = this.props.mode ? false : true;
+    this.IS_INSTALLL = this.props.mode === undefined ? false : true;
```
jeremy-moffitt: just noticed, too many LLL's here like was previously the case in EditServerDetails.js

jeremy-moffitt: also, isnt this the reverse of the intended logic? if this.props.mode is undefined, IS_INSTALL should be true?

src/components/EditServerDetails.js
```diff
@@ -26,7 +26,7 @@ class EditServerDetails extends Component {
   constructor(props) {
     super(props);
 
-    this.IS_INSTALL = this.props.mode ? false : true;
+    this.IS_INSTALL = this.props.mode === undefined ? false : true;
```
jeremy-moffitt: I think this logic is the reverse of what you're intending, will pull it down and try it out

jeremy-moffitt (changes requested):
the logic is clearer on checking IS_INSTALL, but its backwards with the latest change

guglo:
oops...will update.

jeremy-moffitt (approved):

src/pages/AddServers.js
GarySmith: That doesn't make any sense to me, but I'm not going to fret over a possibly extra check

src/pages/AssignServerRoles.js
```diff
@@ -53,7 +53,7 @@ class AssignServerRoles extends BaseWizardPage {
   constructor(props) {
     super(props);
 
-    this.IS_INSTALLL = this.props.mode === undefined ? false : true;
+    this.IS_INSTALL = this.props.mode === undefined ? true : false;
```
GarySmith: I thought I had commented on this before, but I can't find my comment. Anyhow, I find the whole `mode` property confusing. It appears that a value of `undefined` means it is install mode, and a value of `addserver` (I think) puts it into add mode. It is unclear what will happen with any other value. I would suggest using a boolean value instead of `mode` like `installMode` that is more meaningful and avoids having to create extra variable like `this.IS_INSTALL` to make it meaningful. I'd rather not hold up this change for that improvement, so I'm going to mark it as approved, but I'd really like to see this addressed in a followup change.

GarySmith (approved):

GarySmith:
I thought I had commented on this before, but I can't find my comment. Anyhow, I find the whole `mode` property confusing. It appears that a value of `undefined` means it is install mode, and a value of `addserver` (I think) puts it into add mode. It is unclear what will happen with any other value. I would suggest using a boolean value instead of `mode` (say, `installMode`_ that is more meaningful and avoids having to create extra variable like `this.IS_INSTALL` to make sense of it. I'd rather not hold up this change for that improvement, so I'm going to mark it as approved, but I'd really like to see this cleaned up in another change.

guglo:
@GarySmith  thanks Gary, will switch to the boolean installMode to simplify in the next around. Will refactor if we need to deal with more modes in the future.

PR #144, opened 2018-08-27
Merge pike august 27 2018

jeremy-moffitt:
Merge Day2 related changes into stable/pike


guglo:
Generally looks fine. But don't understand why it includes the some small incremental commits ?

jeremy-moffitt:
Unless those commits were squashed at the time their pull-request was merged, then they're in the git log like any other commit... If I squash-merge the lot of this, then I *think* the history would be lost... @GarySmith  is that the case? or should I have squashed this?

guglo:
I see. thanks.

PR #148, opened 2018-08-31
SCRD-3476 added check against deployed servers

ghost:
This is the continue work for 'add compute node'. This
 checkin includes the followings:

1. Added the call to get deployed servers and check against the list
  if the server is in deploy servers, the node is read only and is
  not draggable.
2. If the server in the accordion is missing nic-mapping and
   server-grouop, it will be in red.
3. Added loading error handling in add server page
4. Moved loading error handling to BaseUpdateWizardPage so it
  could be shared. Updated update server page to use shared
  functions.
5. Added loading mask during loading in the add server  page
6. Simplified the 3 ways this.props.mode to use global boolean
isUpdateMode

This UI change depends on ardana-service change
https://gerrit.suse.provo.cloud//#/c/4643

### DIFF
--- a/src/InstallWizard.js
+++ b/src/InstallWizard.js
@@ -23,7 +23,7 @@ import { fetchJson, postJson, deleteJson } from './utils/RestUtils.js';
 
 // global states vars used in installation and update
 const COMMON_GLOBAL_STATES_VARS = [
-  'playbookStatus', 'model', 'connectionInfo', 'deployConfig'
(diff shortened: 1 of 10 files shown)


jeremy-moffitt (approved):
this worked ok for me, and the code looks fine. took me a minute to figure out how/where this.IS_UPDATE was being set because it was already being set, but I see it now :)

PR #149, opened 2018-08-31
SCRD-3490 added loadingmask

ghost:


jeremy-moffitt (commented):
code looks fine, is there an easy way to test this? on my local env the pages comes right up (assuming I'm looking at the right page, maybe I'm not)

guglo:
This is what I did for testing: 
click around , like click servers menu , click any pages on Topology and then click servers menu, you should be able to see the loadingmask on the servers-summary page. it goes fast though.

PR #152, opened 2018-09-05
Add network topology tab

GarySmith:
Add network tab to the topology page.
Add support for passing react components to the translate function.
Add additional Sort functions to facilitate sorting objects and arrays
  of objects.


GarySmith:
The subject of the last commit should say "control plane tab", not "control panel tab". But since we are going to squash and merge, that should all just disappear when the pull request is accepted.

src/localization/localize.js
```diff
@@ -51,11 +52,37 @@ else { //default
...
-  // The join() corrects this by joining the elements into a single string.
+
+  const hasReact = args && args.some(arg => React.isValidElement(arg));
+
   try {
-    return strings.formatString(strings[key], ...args).join('');
+    // React components cannot be handled with String.formatString, so if any
+    // args are react components, then manually parse the format string and
+    // build up the result with
+    if (hasReact) {
```
jeremy-moffitt: is there a good place to test this? it looks good, but I'd like to try out both versions of the fork... does the topology tab do that?

GarySmith: Yes, this is used by the network tab in the topology page, in line 211 of Networks.js. It is used when rendering the Network Tags entries in the first column (Network Group). To see an example of this, use the mid-scale-kvm model (via the install UI, which you can also reach by using localhost:3000?start=installer). Then when you go to http://localhost:3000/#/topology/networks#MANAGEMENT (the Management network network), you can see under the Networks Tags: an entry for the Control Plane. The hyperlink there to 'control-plane-1' is generated through this function

PR #156, opened 2018-09-10
Add servers topology tab

GarySmith:
Add servers tab to the topology page.
Implement styling changes to clarify relationships, making the pages
  more similar to the original versions from the config processor.


GarySmith:
I thought it was an improvement, but I can limit the scope

jeremy-moffitt:
> I thought it was an improvement, but I can limit the scope

the lines probably are an improvement, but the tables not spanning the width of the page (leaving whitespace to the right) looks a bit cramped

PR #157, opened 2018-09-10
SCRD-3478 Update OpenStack and Ardana Packages pages with new API

jgravel:
Updated the OpenStack Packages page and the SUSE OpenStack Cloud Packages page with the new API. This change depends on these backend changes which implement the new API: https://gerrit.suse.provo.cloud/#/c/4708 and https://gerrit.suse.provo.cloud/#/c/4709. Please do not merge this change until the backend changes were merged.


jeremy-moffitt:
works well, looks like this needs to be merged basically as soon as the backend change is merged as they only work together and either one alone breaks the page

PR #167, opened 2018-09-20
SCRD-3486 Add Details modal for service info tab

jgravel:
The Details modal shows information about each service on the Service Information tab. The modal can be launched from the 'Details' menu off of each service row. 

The backend code with Monasca's get alarm count API is available on Oak. To test this change, you need to set your shimurl in config.dev.js to 'http://192.168.200.10:8081'.


jgravel:
They should be the same. Let me take a look.

jgravel:
I checked with Jack. There will be special cases for Ops Console and Ardana service. I'll update the code.

jgravel:
I updated the code to have the following services on the UI map to the real service names:
1. Ardana -> 'ardana-service'
2. Kronos -> 'python-monasca-log-api'
3. Opsconsole -> 'python-ardana-opsconsole-server'

Please take another look.

PR #170, opened 2018-09-25
SCRD-4708 Update Day0 UI name to Cloud Installer

jgravel:


jgravel:
Added more updates, please take a look again.

PR #171, opened 2018-09-26
Incorporate Servers into Server Groups topology tab

GarySmith:
Incorporate the servers information (details about each server) onto
the exising ServerGroups tab, and rename the resulting tab (and file)
to be just Servers.

Update hyperlinks from other topology pages.


src/pages/topology/Servers.js
```diff
@@ -0,0 +1,353 @@
...
+
+  }
+
+  componentDidMount() {
+
+    getInternalModel()
+      .then((yml) => {
+
+        // Force a re-render if the page is still shown (user may navigate away while waiting)
+        if (this.refs.server_groups)
```
jeremy-moffitt: I'm not sure I understand this section, though it seems to work ok and its not technically new since it existed in the previous version of this file (which was renamed and git doesnt handle that well)

src/pages/topology/TopologyUtils.js
```diff
@@ -24,6 +24,11 @@ function loadInternalModel() {
   return fetchJson('/api/v1/clm/model/cp_internal/CloudModel.yaml');
 }
 
+// Prevent wrapping on hyphens by replacing normal hyphen characters with the non-wrapping-hyphen character
+export function noHyphenWrap(s) {
+  return s.replace(/-/g, '\u2011');
```
jeremy-moffitt: this must have been 'fun' to figure out!

PR #175, opened 2018-09-29
SCRD-3477 Implemented install os for adding compute nodes

ghost:
This commit includes:

1. Implemented the Install OS button to launch processes
   to allow user select newly added compute nodes and
   install SLES OS on te compute nodes.
2. Reused and updated SelectServerToProvision to be able to
   handle isUpdateMode
2. Minor fixes to not show progress bar when only one page
   assembled.
3. Minor styling fix to for showing complete server list

### DIFF
--- a/src/pages/AddServers.js
+++ b/src/pages/AddServers.js
@@ -88,44 +88,31 @@ class AddServers extends BaseUpdateWizardPage {
   }
 
   assembleInstallProcessPages = () => {
-    let pages = [];
 
-    pages.push({
+    return [{
       name: 'SelectInstallOS',
       component: AddServersPages.SelectInstallOS
-    });
-
-    pages.push({
+    }, {
       name: 'ProcessInstallOS',
       component: AddServersPages.ProcessInstallOS
-    });
-
-    pages.push({
+    }, {
       name: 'CompleteInstallOS',
       component: AddServersPages.CompleteInstallOS
-    });
-
-    return pages;
+    }];
   }
 
   assembleDeployProcessPages = () => {
-    let pages = [];
 
-    pages.push({
+    return [{
       name: 'PrepareAddServers',
       component: AddServersPages.PrepareAddServers
-    });
-
-    pages.push({
+    }, {
       name: 'DeployAddServers',
       component: AddServersPages.DeployAddServers
-    });
-
-    pages.push({
+    }, {
       name: 'CompleteAddServers',
       component: AddServersPages.CompleteAddServers
-    });
-    return pages;
+    }];
   }
 
   getAddedServerIds = () => {
--- a/src/pages/AddServers.js
+++ b/src/pages/AddServers.js
@@ -87,7 +87,28 @@ class AddServers extends BaseUpdateWizardPage {
       });
   }
 
-  assembleProcessPages = () => {
+  assembleInstallProcessPages = () => {
+    let pages = [];
+
+    pages.push({
+      name: 'SelectInstallOS',
+      component: AddServersPages.SelectInstallOS
+    });
+
+    pages.push({
+      name: 'ProcessInstallOS',
+      component: AddServersPages.ProcessInstallOS
+    });
+
+    pages.push({
+      name: 'CompleteInstallOS',
+      component: AddServersPages.CompleteInstallOS
+    });
+
+    return pages;
+  }
+
+  assembleDeployProcessPages = () => {
     let pages = [];
 
     pages.push({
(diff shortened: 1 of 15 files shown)


guglo:
Tested in dev...still doing testing on provo system

guglo:
Tested on provo system to add compute2

jeremy-moffitt:
needs rebased... also one nitpick functionally, the "next" and "cancel" buttons are both green as if they're primary action buttons on the screen where you pick which servers to install the OS on... 

still need to look at the code... the functionality seems good

src/pages/AddServers/CompleteInstallOS.js
```diff
@@ -0,0 +1,83 @@
...
+import { alphabetically } from '../../utils/Sort.js';
+
+// This is the complete page for installing SLES OS on compute servers
+// It will display server ID, IP, mac-addr and ilo-ip
+class CompleteInstallOS extends BaseUpdateWizardPage {
+  constructor(props) {
+    super(props);
+    this.state = {
+      // loading errors from wizard model or progress loading
+      wizardLoadingErrors: this.props.wizardLoadingErrors,
```
jeremy-moffitt: I'm surprised these cant just be referenced off of the props object, do they have to be part of the state? they arent changed within this class, applies to the other classes that use this as well, maybe I'm just react rusty...

jeremy-moffitt (commented):
I'm confused why some of the props are being copied to the state object (and are only updated when new props come in), is there a reason these can't be directly referenced from the props objects? otherwise just needs rebased and the green cancel button fixed (per above), overall looks good

src/pages/AddServers/CompleteInstallOS.js
guglo: let me try if I can get rid of some code

guglo:
I tried best to clean up some of the unnecessary states and tried the possible testing ....will go another around of clean up and testing to see if I missed anything or some cleanup messed up.

src/pages/AddServers.js
```diff
@@ -87,7 +87,28 @@ class AddServers extends BaseUpdateWizardPage {
       });
   }
 
-  assembleProcessPages = () => {
+  assembleInstallProcessPages = () => {
+    let pages = [];
```
GarySmith: Instead of defining `pages` to be an empty array and using `push` to append to it, it would be simpler to just return a static array, i.e.:
```
return ([{
  name: 'SelectedInstallOS',
  component: AddServersPages.SelectInstallOS
},{
  name: 'ProcessInstallOS',
  component: AddServersPages.ProcessInstallOS
},{
  name: 'CompleteInstallOS',
  component: AddServersPages.CompleteInstallOS
});
```

src/pages/AddServers/CompleteInstallOS.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  renderServerList() {
+    let serverList = this.props.operationProps.selectedToInstallOS.sort((a, b) => {
+      return alphabetically(a['id'], b['id']);
+    }).map((host, idx) => {
+      return (
+        <li key={idx}>{host['id']}
+          <ul className='complete-serverdetails'>
+            <li key={idx + '_1'}> {translate('server.details.ip') + ' ' + host['ip-addr']} </li>
```
GarySmith: This `{idx+'_1'}` construct is unnecessary, since the LI keys only have to be unique within this list, i.e. you can just use `key='1'`

src/pages/AddServers/SelectInstallOS.js
```diff
@@ -0,0 +1,125 @@
...
+      // error msg
+      passphraseError: undefined,
+      // disable the next button when it is not valid
+      isInvalid: true
+    };
+  }
+
+  componentWillReceiveProps(newProps) {
+    let isInvalid =
+      !newProps.operationProps.selectedToInstallOS ||
```
GarySmith: Instead of using this same pattern everywhere of checking that the property exists and that its length is 0 (or not 0), please consider using `isEmpty` from lodash. It would make it much easier to understand, e.g.: 
```
  const isInvalid = 
    isEmpty(newProps.operationProps.selectedToInstallOS) ||
    isEmpty(newProps.operationProps.osInstallPassword) ||
    (isEmpty(newProps.operationProps.sshPassphrase) && newProps.operationProps.sshPassphraseRequired);
```
I also see this in other places in this PR

GarySmith (changes requested):
Mostly small stuff needed

jeremy-moffitt:
I noticed that the minified version of the app on master still hides the add-server screen, we should change that if its still the case with this commit (I havent had a chance to re-test it yet)

src/pages/AddServers.js
guglo: ok will update

src/pages/AddServers/CompleteInstallOS.js
guglo: k will update

src/pages/AddServers/SelectInstallOS.js
guglo: k will update

guglo:
will make available

GarySmith (approved):

PR #179, opened 2018-10-04
Improve replace controller dialog

GarySmith:
Before this change the replace controller dialog had a number of sub-optimal constructs including copying props into state, using class variables to hold state, and so on. The form would also trigger many unnecessary renders.
Removed all class variables and copying of props, and improved the handling of state to minimize renders.


GarySmith:
Note that PR was performed as several smaller commits, and these were kept intact (rather than squashing) to make it a little easier to follow along. Unfortunately the last commit to fix the state handling is still rather large. These should be all squashed during the merge.

PR #181, opened 2018-10-04
SCRD-4773 Merge changes from master to stable/pike oct 04 2018

jeremy-moffitt:


jeremy-moffitt:
It's odd to me that this merge commit had so many conflicts, which result in this commit appearing to have a ton of changes that actually were sourced from other commits... github is weird about merge commits

PR #182, opened 2018-10-05
SCRD-3477 fixed issue related to check addresses conflict

ghost:
This commit includes:

1. Updated isInstallable logic to be specific for the installation
   where it needs to make sure have new servers, at least one server
   has all the info to install and they are not conflicts with existing
   ones in the model.

2. For deploy, it should not care about empty mac-addr or ilo-ip

### DIFF
--- a/src/pages/AddServers.js
+++ b/src/pages/AddServers.js
@@ -164,7 +164,6 @@ class AddServers extends BaseUpdateWizardPage {
   }
 
   hasInvalidNewServers = (checkForInstall) => {
-    let isServersInvalid = false;
     let allSevers = this.state.model.getIn(['inputModel','servers']).toJS();
     let deployedServerIds =
       this.state.deployedServers ?  this.state.deployedServers.map(server => server.id) : [];
@@ -178,9 +177,8 @@ class AddServers extends BaseUpdateWizardPage {
     // check if newly added servers have addresses conflicts with any deployed servers
     for (let i = 0; i < newServers.length; i++) {
       let newServer = newServers[i];
-      isServersInvalid = hasConflictAddresses(newServer, modelDeployedServers);
-      if (isServersInvalid) {
-        return isServersInvalid;
+      if (hasConflictAddresses(newServer, modelDeployedServers)) {
+        return true;
       }
     }
 
@@ -190,29 +188,24 @@ class AddServers extends BaseUpdateWizardPage {
       let hasOne = newServers.some(server =>
         !isEmpty(server['mac-addr']) && !isEmpty(server['ilo-ip']) &&
         !isEmpty(server['ilo-user']) && !isEmpty(server['ilo-password']));
-      isServersInvalid = !hasOne;
-      if(isServersInvalid) {
-        return isServersInvalid;
+      if(!hasOne) {
+        return true;
       }
     }
 
     // check if have duplicates within the newly added servers
     let addresses = newServers.map(server => server['mac-addr']);
-    isServersInvalid = this.hasDuplicates(addresses);
-    if(isServersInvalid) {
-      return isServersInvalid;
+    if(this.hasDuplicates(addresses)) {
+      return true;
     }
 
     addresses = newServers.map(server => server['ip-addr']);
-    isServersInvalid = this.hasDuplicates(addresses);
-    if(isServersInvalid) {
-      return isServersInvalid;
+    if(this.hasDuplicates(addresses)) {
+      return true;
     }
 
     addresses = newServers.map(server => server['ilo-ip']);
-    isServersInvalid = this.hasDuplicates(addresses);
-
-    return isServersInvalid;
+    return this.hasDuplicates(addresses);
   }
 
   installOS = () => {
--- a/src/pages/AddServers.js
+++ b/src/pages/AddServers.js
@@ -14,6 +14,7 @@
 **/
 
 import React from 'react';
+import { isEmpty } from 'lodash';
 import { AddServersPages } from './AddServers/AddServersPages.js';
 import AssignServerRoles from './AssignServerRoles.js';
 import BaseUpdateWizardPage from './BaseUpdateWizardPage.js';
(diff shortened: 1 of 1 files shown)


src/pages/AddServers.js
```diff
@@ -150,11 +151,20 @@ class AddServers extends BaseUpdateWizardPage {
   // values, then set size is the same as the array list length,
   // otherwise the list contains duplicate values.
   hasDuplicates = (arrayList) => {
-    return (new Set(arrayList)).size !== arrayList.length;
+    if(isEmpty(arrayList)) {
```
GarySmith: This check is unnecessary. arrayList is the result of a map() function, so it will always be a valid array, and the rest of the logic works correctly when the array is empty

src/pages/AddServers.js
```diff
@@ -150,11 +151,20 @@ class AddServers extends BaseUpdateWizardPage {
...
+      return false;
+    }
+    // filter out empty items
+    let cleanList = arrayList.filter(item => !isEmpty(item));
+    if(isEmpty(cleanList)) {
```
GarySmith: Again, remove this check. It is unnecessary.

src/pages/AddServers.js
```diff
@@ -150,11 +151,20 @@ class AddServers extends BaseUpdateWizardPage {
...
+      return false;
+    }
+
+    return (new Set(cleanList)).size !== cleanList.length;
```
GarySmith: A comment here would be helpful. It is not immediately obvious that the point of creating a Set is to remove duplicates.

src/pages/AddServers.js
```diff
@@ -150,11 +151,20 @@ class AddServers extends BaseUpdateWizardPage {
...
   }
 
-  hasAddressesConflicts = () => {
-    let hasConflicts = false;
```
GarySmith: instead of having this variable `hasInvalid`, the logic would be simpler if you just immediately return when you spot a violation. Also, it would be simpler to understand if this function were in the positive, i.e. `isServerValid`, and you return `true` when it is valid and `false` otherwise.

GarySmith (changes requested):

src/pages/AddServers.js
guglo: that is true, will update

src/pages/AddServers.js
guglo: this is actual fix point where I can have array list contain like 2 empty item...if arrayList has [ undefined, undefined], cleanList is empty, so this check is necessary. This actually fixed the problem where has 2 server items , both of them have empty mac-addr...

src/pages/AddServers.js
guglo: comments are at  the top of the method..I can move it down

src/pages/AddServers.js
```diff
@@ -168,29 +178,41 @@ class AddServers extends BaseUpdateWizardPage {
     // check if newly added servers have addresses conflicts with any deployed servers
     for (let i = 0; i < newServers.length; i++) {
       let newServer = newServers[i];
-      hasConflicts = hasConflictAddresses(newServer, modelDeployedServers);
-      if (hasConflicts) {
-        break;
+      isServersInvalid = hasConflictAddresses(newServer, modelDeployedServers);
```
GarySmith: No, you still don't need the variable or any of the checks of its value. This code, and all of the other checks below, should look something like:
```
if (hasConflictAddresses(newServer, modelDeployedServers))
  return true;
```

src/pages/AddServers.js
GarySmith: When functions are coded to check whether something is valid or invalid, it is generally easier to understand them if true represent the valid case and false represents invalid.

src/pages/AddServers.js
```diff
@@ -144,17 +145,26 @@ class AddServers extends BaseUpdateWizardPage {
...
+    // unique values. If list doesn't contain duplicate
+    // values, then set size is the same as the list length,
+    // otherwise the list contains duplicate values.
+    return (new Set(cleanList)).size !== cleanList.length;
   }
 
-  hasAddressesConflicts = () => {
-    let hasConflicts = false;
+  hasInvalidNewServers = (checkForInstall) => {
+    let isServersInvalid = false;
```
GarySmith: As discussed before (and below), this variable should be removed

src/pages/AddServers.js
```diff
@@ -222,7 +244,20 @@ class AddServers extends BaseUpdateWizardPage {
   }
 
   isInstallable = () => {
-    return this.isDeployable();
+    if(this.state.model && this.state.model.size > 0) {
```
GarySmith: I don't understand why the size check is here. When would you ever have a model which is not `undefined` but which is empty?

src/pages/AddServers.js
GarySmith: Thanks, that is better.

src/pages/AddServers.js
GarySmith: That doesn't make sense. If `cleanList` was empty, then the check on the next line (163) which compares the size of the set with the size of the list would still return false.

GarySmith (changes requested):

src/pages/AddServers.js
guglo: at certain point during testing...noticed model.size could be 0 sometimes.... Added it to void problems.

src/pages/AddServers.js
guglo: updated with valid is true invalid is false now.

GarySmith (approved):
Thanks for making those changes.

jeremy-moffitt:
so I tried adding 2 new computes simultaneously, it goes through the workflow ok, but when I click "close" at the end to reload the servers page, there is a load mask that never goes away... I'm going to try again to see if the behavior repeats

guglo:
when that happens...if watch the request to get deployed servers in ardana-service...sometimes  ardana-service doesn't come back...similar to config-processor issue...

PR #184, opened 2018-10-06
SCRD-5295 moved update processes out of InstallWizard

ghost:
This commit includes:
1. Moved update processes out of InstallWizard
2. Moved global states and persiste states to constants so
   they can be shared.
3. Moved some processes in InstallWizard into function so it can
   be overwritten.
4. Created UpdateWizard, as first step, it extends InstallWizard so
   core functions in InstallWizard can be shared.
5. Updated AddServer page and UpdateServer landing page to use UpdateWizard
6. Updated other components accordingly because of the move
7. Replaced componentWillReceiveProps with componentDidUpdate
8. Reduced unnecessary states.

### DIFF
--- a/src/InstallWizard.js
+++ b/src/InstallWizard.js
@@ -21,38 +21,14 @@ import { LoadingMask } from './components/LoadingMask.js';
 import { Map, fromJS } from 'immutable';
 import { fetchJson, postJson, deleteJson } from './utils/RestUtils.js';
 
-// global states vars used in installation and update
-const COMMON_GLOBAL_STATES_VARS = [
-  'playbookStatus', 'model', 'connectionInfo', 'deployConfig', 'isUpdateMode'
-];
-
-// global states vars used in update
-const UPDATE_GLOBAL_STATES_VARS = [
-  'currentMenuName', 'processMenuName', 'processOperation', 'steps', 'pages', 'operationProps',
-  'wizardLoadingErrors', 'wizardLoading'
-];
-
-// global persisted states vars used in installation and update
-const COMMON_PERSISTED_STATES_VARS = [
-  'currentStep', 'steps', 'playbookStatus', 'connectionInfo', 'deployConfig'
-];
-
-// global persisted states vars used in update
-const UPDATE_PERSISTED_STATES_VARS = [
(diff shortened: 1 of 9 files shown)


src/InstallWizard.js
```diff
@@ -20,40 +20,19 @@ import WizardProgress from './components/WizardProgress';
...
-// global persisted states vars used in installation and update
-const COMMON_PERSISTED_STATES_VARS = [
-  'currentStep', 'steps', 'playbookStatus', 'connectionInfo', 'deployConfig'
-];
-
-// global persisted states vars used in update
-const UPDATE_PERSISTED_STATES_VARS = [
-  'currentMenuName', 'processMenuName', 'processOperation', 'operationProps'
-];
+import {
```
GarySmith: I would prefer that these arrays not be put into `constants.js`, since they are very specific to the internal workings of this class and should not be exposed elsewhere. Furthermore, I don't even think they need to be constants since  they are each used only one line in the class (in the constructor).

src/UpdateWizard.js
```diff
@@ -0,0 +1,285 @@
...
+      // pages are not persisted
+      pages: undefined,
+
+      // operationProps record all the bits that need to run to update
+      // TODO how to deal with install password
+      operationProps: undefined,
+    };
+
+    // Indicate which of the above state variables are passed to wizard pages and can be set by them
+    this.globalStateVars =
```
GarySmith: Instead of having all of these global constants (which expose inner details of these classes too broadly), it would be better to do:
`this.globalStateVars = this.globalStateVars.concat(['currentMenuName, 'processMenuName' ...]);`

GarySmith (changes requested):
This is a *great* improvement over the previous version. Thanks so much for doing this. Just one small change suggested here.

src/InstallWizard.js
guglo: k, will look into it.

src/UpdateWizard.js
guglo: k, will look into it.

guglo:
need more testing.

GarySmith (approved):
LGTM. Lots of good cleanup here.

PR #185, opened 2018-10-08
SCRD-3476 fixed display hostname

ghost:
The hostname to set --limit when deploying is different from the hostname
displayed in horizon or opsconsole.

Fixed the complete page to use display hostname so they match with
the hostnames in horizon or opsconsole.


GarySmith (approved):
LGTM (Caveat -- did not test this)

PR #186, opened 2018-10-08
SCRD-3497 Fix field validation in replace server dialog

GarySmith:
The field validation in the replace server dialog had three problems:
- It did not exclude in-use servers from the pulldown list of
  available servers.
- The duplicate check should have flagged only those values used in
  a server in the model
- The duplicate IP address check should have checked against IP
  addresses used either as the main server address or its ILO address

In the process of improving the parameters to the validator I found
that the existing mechanism of checking for duplicates required very
tight coupling between the validator function and the field being
validated: specifically, the validator required that the input be
passed with two hardcoded properties, each of which needed to contain
a very specific structure.  Rather than propagating that further,
established a more general validation function for checking for
duplicates and for checking multiple criteria.

### DIFF
--- a/src/components/CollapsibleTable.js
+++ b/src/components/CollapsibleTable.js
@@ -310,7 +310,16 @@ class CollapsibleTable extends Component {
   renderReplaceServerModal() {
     let title = translate('server.replace.heading', this.state.activeRowData.id);
     let newProps = { ...this.props };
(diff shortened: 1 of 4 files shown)


GarySmith:
Note that I rebased against master and re-pushed

PR #188, opened 2018-10-09
Modernize UpdateServers

GarySmith:
Removed calls to two deprecated ReactJS lifecycle methods,
componentWillMount and componentWillReceiveProps, and replaced with
alternatives.  Make only a single call to the installer service to
retrieve all manually and automatically discovered servers . Remove
storing the input model in state since it is already being passed in
props.  Handle missing roles more simply and correctly.


jeremy-moffitt:
is there an SCRD we can associate with the commit message for this?

PR #189, opened 2018-10-09
SCRD-3489 Add Update Service Config - Part 2

jgravel:
Added the update functionality to the 'Update' button on the page. Note that when I tested in mock mode, the update ran all the way through. But when I tested with Oak's backend, the log message area showed messages of the update process until it hit the 'stop haproxy' command, then stopped showing any more messages. The update process however still proceeded in the backend. This issue can be solved in a separate pull request.


jeremy-moffitt:
I wonder if instead of "back" we should show "Close" or "Done" or something when its finished... I guess its not a modal, so maybe "Close" doesn't make sense... hmm I'm not sure... overall looks good

PR #194, opened 2018-10-16
SCRD-3497 Refactor to enable replece controller

GarySmith:
Refactor CollapsibleTable so that the logic for
replacing the controller can be put in a single place -- UpdateServer,
rather than spreading the logic between several components.
Remove several unused functions and renamed variables for clarity.
Remove unnecessary state variables
Simplify showing and hiding of dialog boxes and context menus


GarySmith:
Rebased against master

src/pages/ReplaceServer/PrepareReplace.js
```diff
@@ -31,6 +31,7 @@ const PLAYBOOK_STEPS = [
     label: translate('deploy.progress.ready-deployment'),
     playbooks: ['ready-deployment.yml']
   },
+  // TODO: Remove this. It is just confusing
```
jeremy-moffitt: It's not clear what this comment in referring to

GarySmith: something I removed but forgot to remove the comment. Will update

GarySmith: actually, it's a valid comment that needs clarification. This code is confusing in that it runs the config processor playbook, then ready deployment, then the installui-pre-deployment (PRE_DEPLOYMENT_PLAYBOOK) -- the confusing part is that this last playbook just runs those first two playbooks again.

GarySmith: I think there is some code somewhere that causes them to be run once instead of twice, but it is really confusing to define essentially the same set of playbooks twice.

GarySmith: Added a commit to clarify that comment

GarySmith:
One more note: when squash/merging this commit, please fix the commit message summary line to say 'replace' rather than 'replece'

src/components/CollapsibleTable.js
```diff
@@ -221,21 +124,67 @@ class CollapsibleTable extends Component {
...
+            key: 'common.replace',
+            action: this.props.replaceServer,
+            callbackData: row
+          });
+        }
+      }
+    } else {
+      // not compute node
+      if(!isProduction()) {
+        items.push({
```
guglo: you want to have !this.props.processOperation as condition here as well

GarySmith: ok, i'll add it. You didn't have it the version that you committed to master, so I figured you wanted it that way for some reason.

GarySmith: Actually, let's add that in as a separate change, since this was not introduced by this PR

GarySmith:
Yes, SCRD-5340 is the bug for the replace dialog not updating the model, which is also not part of this PR. I have to admit that for repeated testing, it is nice to not update the model, so my desire to fix that in the short term is low

PR #195, opened 2018-10-16
 SCRD-3497 Disallow replacing shared controller/deployer

GarySmith:
Detect the situation where the controller that has been requested to be
replaced is also hosting the deployer, and prompt the user to perform
this manually.  Replacing a shared node is far more complex and is not
included in this user story.

### DIFF
--- a/src/pages/UpdateServers.js
+++ b/src/pages/UpdateServers.js
@@ -26,7 +26,7 @@ import { MODEL_SERVER_PROPS_ALL, REPLACE_SERVER_PROPS } from '../utils/constants
 import { updateServersInModel, getMergedServer } from '../utils/ModelUtils.js';
 import { fetchJson, putJson } from '../utils/RestUtils.js';
 import ReplaceServerDetails from '../components/ReplaceServerDetails.js';
-import { BaseInputModal } from '../components/Modals.js';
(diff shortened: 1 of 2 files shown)


GarySmith:
Please don't merge this until I can rebase it.

GarySmith:
Rebase complete

src/pages/UpdateServers.js
```diff
@@ -191,13 +193,22 @@ class UpdateServers extends BaseUpdateWizardPage {
     return (<div className='notification-message-container'>{msgList}</div>);
   }
 
-  handleShowMenuReplaceServer = (server) => {
-    this.setState({
-      showReplaceModal: true,
-      serverToReplace: server
-    });
+  checkSharedControllerDeployer = (server) => {
+    fetchJson('api/v1/ips')
```
jeremy-moffitt: its a little unclear what the business logic is here.. I assume its getting a list of ips of the current server (which would be the deployer since the UI is hosted there) and then checking it against the IP of the server being replaced ... I guess I'm fuzzy on api/v1/ips...

GarySmith: Yes, that's exactly what it does. FWIW, SelectServersToProvision also has similar logic to avoid doing an OS provision of a node if it is the one hosting the deplyer.

PR #196, opened 2018-10-17
Simplify components and convert to stateless functions

GarySmith:
Remove unnecessary state and convert to stateless functional components

### DIFF
--- a/src/components/InputLine.js
+++ b/src/components/InputLine.js
@@ -12,25 +12,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
-import React, { Component } from 'react';
+import React from 'react';
 import { ValidatingInput } from './ValidatingInput.js';
 import { translate } from '../localization/localize.js';
 
-export class InputLine extends Component {
-  render() {
-    let labelStr = this.props.label ? translate(this.props.label) : '';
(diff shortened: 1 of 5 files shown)


guglo:
Noticed regression forListDropDown.: When AddServer manually Nic-mapping and role select doesn't seem to select.

GarySmith:
Grrr. It works for edit mode but not add mode. The problem is that a number of variables in this ServersAddedManually belong in state and are not there (those that the user has entered and should trigger re-rendering when changed). I'll fix this dialog too

GarySmith:
I found a bug in the refresh logic that supressed updates when in add mode and submitted that. Cleaning up the state can wait for another time.

guglo:
@GarySmith  Not sure it is ready, but I tried the latest change, AddServerManually modal now allows me select now, but the buttons don't get enabled.

GarySmith:
OK, I tested this dialog in many ways, and it works correctly in every way I tried. @guglo can you break it?

guglo:
@GarySmith  ok will give it a try :)

GarySmith:
Thanks for testing it and spotting the lint warnings (my auto-lint-checking plugin in vim was temporarily disabled and I missed those).

PR #197, opened 2018-10-18
SCRD-3497 Implement check for power off state

GarySmith:


src/pages/UpdateServers.js
```diff
@@ -193,21 +193,50 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+                  errorMessages: prev.errorMessages.concat(msg),
+                  loading: false
+                }));
+              }
+            });
         }
       });
   }
 
+  enforcePowerOff = (server) => {
```
jeremy-moffitt: not sure what this is doing here....

GarySmith:
There appears to be no way to suppress these console messages, at least not in chrome. For example, this indicates it is an outstanding bug https://bugs.chromium.org/p/chromium/issues/detail?id=124534 . I added a clarifying log message to avoid confusion

PR #199, opened 2018-10-25
SCRD-3497 Support running non-playbooks with PlaybookProgress

GarySmith:
Add support for supplying an arbitrary action instead of only launching
ansible playbooks. The action is a function that returns a promise,
which enables it to be used for making API calls to the back end as well
as any other call.
Replaced logic that used isUpdateMode to indicate that the playbook
array was to contain objects (instead of just names) with logic that
inspected the array and made the decision based on the contents.
Added documentation for the props passed in to PlaybookProcess.
Added an example function in PrepareReplace.

### DIFF
--- a/src/components/PlaybookProcess.js
+++ b/src/components/PlaybookProcess.js
@@ -142,6 +142,47 @@ class PlaybookProgress extends Component {
       showConfirmationDlg: false,        // show the confirmation dialog
       closeStatusPlaybookModal: false,   // set to true to close the Run Status Playbook modal
     };
+
+    // This component expects the following values to be passed in props:
+    //   updatePageStatus - function to be called whenever a playbook starts, completes, or fails
+    //                      Used by callers to show an error dialog or enable navigation buttons
+    //                      (sometimes an empty function is passed)
+    //
+    //   updateGlobalState - function (usually in the installWizard) for updating the global state. Used in this class
+    //                       for updating playbookStatus, which in turn is received separately as a prop.
+    //
+    //   playbookStatus - array of playbooks, playIdsm, and statuses.  Used to track status of all playbooks
+    //
+    //   playbooks - EITHER:
+    //       - array of playbook names to invoke (without the .yml suffix)
+    //          OR
+    //       - array of objects, containing:
+    //            name - name of the playbook
+    //            payload - payload specific to *this* playbook. This enables passing a specific payload
+    //                      to this playbook
+    //            action - when this is supplied, it should be a function that returns a promise, and it will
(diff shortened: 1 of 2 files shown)


GarySmith:
Note that this enhancement is needed for the stories for replacing compute nodes and controller nodes, since some portions of these processes require mixing playbooks and other commands, and it is much easier for the user (and the developer!) to be able to perform all of this within one page, rather than having to introduce extra pages just to handle that case.

guglo:
@GarySmith  Thanks for the changes! 
I tested day0 os installation and deployment and tested day2 add server os installation and deployment with refresh and without refresh in the dev env, they work very well. 
I tested day2 replaceserver without refresh it is fine. Tested with refresh, the check version has no status after refresh. Just wondering, if you plan to tackle refresh API process call in this commit, or next commit?

GarySmith:
Thanks for catching that. I'll look into it.

GarySmith:
It appears that the code for refresh in PlaybookProgress also needs some logic for knowing that these non-playbook actions are completed even when there is no log available for them to replay. I'll address that in this PR.

guglo:
@GarySmith. Now the action step shows green with checkmark  when refresh. 

 A couple of issues found: 
1. When the step api call failed, it doesn't  show as the step failed...but show as a whole page banner.  I fixed it in PrepareAddServer when I tested add server, but didn't fix in replace server prepare.  At the PrepreReplace, line 141, remove !cancel condition to show step progress. Once I removed that the progress steps with log box shows up + banner when failed. But no errors show up in the log box for action step failure. Need output errors there.

2. When it is successful or failed,   refresh ..it shows it is successful/failed step,  but the logs for action call response/errors are gone,  thinking might want record the results/errors in the progress.json as well for action api calls. 

3. Also refer to PrepareAddServer for the condition to commit model code to avoid unnecessary commit calls when refresh. I fixed it in PrepareAddServer,  didn't fix ReplacePrepare.  Sorry about that. You don't need to fix it in this pr, but don't forget.

GarySmith:
1. I'll make that change
2. The challenge here is that with playbooks, the ardana-service is capturing the log of the command as it runs and forwards it back to the client. In the case of running non-playbooks, there is no such log generated on the back end. One way to make this work would be to add some sort of log service that the UI would write to every time it logs to the screen, and have it replayed in the event of a refresh. But I think that is a lot of overhead and development effort just to re-create the log messages from API commands in the event of a refresh, especially when the steps on the left are correct. I created SCRD-5568 to capture this as a story.
3. Perhaps you can either make this change or I can do it separately from this PR, since it is not something introduced by this PR.

PR #200, opened 2018-10-27
SCRD-3501 Update to add compute node in replace details

ghost:
This commit includes:

1. Added section to have inputs for new compute node in repalce server
   details modal. Replace compute node will add compute node first.

2. Fixed issue when replace non-compute node saves the replace
   information to the model.

### DIFF
--- a/src/components/ReplaceServerDetails.js
+++ b/src/components/ReplaceServerDetails.js
@@ -14,18 +14,26 @@
 **/
 
 import React, { Component } from 'react';
+import { isEmpty } from 'lodash';
 import { translate } from '../localization/localize.js';
 import { ActionButton } from '../components/Buttons.js';
 import { InputLine } from '../components/InputLine.js';
 import { ListDropdown } from '../components/ListDropdown.js';
-import { IpV4AddressValidator, MacAddressValidator, createExcludesValidator, chainValidators }
+import {
+  IpV4AddressValidator, MacAddressValidator, UniqueIdValidator,
+  createExcludesValidator, chainValidators }
   from '../utils/InputValidators.js';
-import { genUID, maskPassword } from '../utils/ModelUtils.js';
+import {
+  genUID, maskPassword, getNicMappings, getServerGroups, getAllOtherServerIds }
+  from '../utils/ModelUtils.js';
 import HelpText from '../components/HelpText.js';
 import { Map, List } from 'immutable';
 import {fetchJson} from '../utils/RestUtils.js';
 
 const Fragment = React.Fragment;
+
+const MAC_IPMI_INPUT_NAMES = ['ilo-ip', 'ilo-user', 'ilo-password', 'mac-addr'];
+const SERVER_INFO_INPUT_NAMES = ['id', 'ip-addr', 'nic-mapping', 'server-group'];
 
 class ReplaceServerDetails extends Component {
   constructor(props) {
@@ -35,29 +43,20 @@ class ReplaceServerDetails extends Component {
       isInstallOsSelected: false,
       isWipeDiskSelected: false,
       isUseAvailServersSelected: false,
-      inputValue: Map({
-        'ilo-user': '',
-        'ilo-password': '',
-        'ilo-ip': '',
-        'mac-addr': '',
-      }),
+      inputValue: this.initInputs(),
 
       selectedServerId: undefined,
       osInstallUsername: undefined,
       osInstallPassword: undefined,
 
-      isValid: Map({
-        'ilo-user': undefined,
-        'ilo-password': undefined,
-        'ilo-ip': undefined,
-        'mac-addr': undefined,
-      }),
+      isValid: this.initInputsValid(),
       isOsInstallPasswordValid: undefined,
+      nicMappings: getNicMappings(props.model),
+      serverGroups: getServerGroups(props.model)
     };
   }
 
-  // This should be componentDidMOunt, since reactjs has deprecated ComponentWillMount
-  componentWillMount() {
+  componentDidMount() {
     fetchJson('/api/v1/clm/user')
       .then(responseData => {
         this.setState({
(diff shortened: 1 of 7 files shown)


src/components/ReplaceServerDetails.js
```diff
@@ -66,22 +65,81 @@ class ReplaceServerDetails extends Component {
...
   isFormInputValid = () => {
 
     // The form is valid if all of the fields are valid.  The OS install password only has
     // to be valid when the OS Install checkbox is selected
-    return this.state.isValid.every((value) => value === true) &&
+    return this.isServerInputsValid() &&
       (this.state.isOsInstallPasswordValid || !this.state.isInstallOsSelected);
   }
 
   handleDone = () => {
```
GarySmith: The function has some problems with naming and cohesion that I am really struggling to understand, and now that we are making substantive changes to this, I'd like to see it fixed if possible:
1. The name `data` is too generic to make any sense of, and it needs a better name. It it being passed to a function in the caller that is expecting a parameter of `server`, so I think server would be a much better name here.
2. `theProps` is similarly so generic that it is impossible to make sense out of. Worse yet, the caller seems to be using it in a variety of ways -- one function is looking for a `wipeDisk` field and ignoring the rest, while other code is doing other things. It is really difficult to understand how this is going to be populated, and it is not clear why all of these fields have to be combined together in ReplaceServerDetails into one big object. It would be far more easy to understand to call `this.props.doneAction` and explicitly list each parameter instead of combining them, i.e.
this.props.doneAction(server, this.state.isWipeDiskSelected, this.state.isInstallOsSelected, ...)
and then let the caller, `UpdateServers`, use them or package them according to its needs.

src/components/ReplaceServerDetails.js
```diff
@@ -92,19 +150,26 @@ class ReplaceServerDetails extends Component {
...
+      let selServer =
+        this.props.availableServers.find(server => server.id === this.state.selectedServerId);
+      data['uid'] = selServer['uid'];
     }
     else {
       // user input new mac-addr and ilo info
-      // generate a new uid
-      data.uid = genUID();
+      // generate a new uid, treat it as manual added server
+      data['uid'] = genUID('manul');
```
GarySmith: This dialog box should just be returning what the selectedServerId is, and the caller should be generatign a new ID if one is necessary. (BTW `manual` is misspelled here)

src/components/ReplaceServerDetails.js
```diff
@@ -92,19 +150,26 @@ class ReplaceServerDetails extends Component {
...
+    }
+
+    // will record the old compute host information so we can process removing
+    // the compute host later
+    if(this.isCompute()) {
```
GarySmith: This does not look right. The caller is the one supplying these props values, so we should not be returning them back to the caller. The caller can build up this structure if needed.

src/components/ReplaceServerDetails.js
```diff
@@ -66,22 +65,81 @@ class ReplaceServerDetails extends Component {
...
-    data['ilo-ip'] = this.state.inputValue.get('ilo-ip');
-    data['ilo-user'] = this.state.inputValue.get('ilo-user');
-    data['ilo-password'] = this.state.inputValue.get('ilo-password');
+    // if it is compute node, will take all user inputs
+    // for the new compute node
+    if(this.isCompute()) {
+      MODEL_SERVER_PROPS.forEach(input_name => {
+        data[input_name] = this.state.inputValue.get(input_name);
+      });
+      data['role'] = this.props.data['role'];
```
GarySmith: This does not look right. The caller is supplying this data already, so we should not be turning around and feeding it right back. Let the caller insert it into this object if it is needed.

src/components/ReplaceServerDetails.js
```diff
@@ -66,22 +65,81 @@ class ReplaceServerDetails extends Component {
...
+    }
+    else { // if it is non-compute node, only replace the mac-addr and ipmi info
+      data = Object.assign({}, this.props.data);
```
GarySmith: Again, we should not be feeding back data to the caller that it supplied to us.

src/pages/UpdateServers.js
```diff
@@ -52,14 +54,15 @@ class UpdateServers extends BaseUpdateWizardPage {
 
       showSharedWarning: false,
 
-      serverToReplace: undefined,
```
GarySmith: FYI, a trailing comma in object lists is officially supported in JavaScript now, so you don't have to remove these, but I won't object if you do.

src/components/ReplaceServerDetails.js
```diff
@@ -188,6 +248,33 @@ class ReplaceServerDetails extends Component {
     }
   }
 
+  isCompute = () => {
```
GarySmith: Please move this function into `ModelUtils`. It is clearly something that would be useful to have in several places in the code. It would be better to name it `isComputeNode` to make it clearer what we are talking about.

src/pages/UpdateServers.js
```diff
@@ -115,16 +118,36 @@ class UpdateServers extends BaseUpdateWizardPage {
     this.setState(prevState => ({'expandedGroup': prevState.expandedGroup.concat(groupName)}));
   }
 
-  updateServerForReplaceServer = (server) => {
+  getReplaceProps = (isCompute) => {
```
GarySmith: This should be calling `isComputeNode` from `ModelUtils`.

src/pages/UpdateServers.js
```diff
@@ -115,16 +118,36 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+    // to saved servers
+    else if(isCompute) {
+      postJson('/api/v1/server', server)
+        .catch(error => {
+          let msg = translate('server.save.error', error.toString());
+          this.setState(prev => ({ errorMessages: prev.errorMessages.concat(msg)}));
+        });
+    }
+    // for non-compute host replacement, if user added info manually, won't
+    // save to saved servers
```
GarySmith: Why aren't controller nodes stored there as well?

src/pages/UpdateServers.js
```diff
@@ -240,8 +277,8 @@ class UpdateServers extends BaseUpdateWizardPage {
     this.setState({showReplaceModal: false, serverToReplace: undefined});
   }
 
-  handleDoneReplaceServer = (server, wipeDisk, installOS) => {
-    this.replaceServer(server, wipeDisk, installOS);
+  handleDoneReplaceServer = (server, theProps) => {
+    this.replaceServer(server, theProps);
```
GarySmith: the old version was better because it was more explicit about what was really needed.

src/styles/components/replaceserverdetails.less
```diff
@@ -25,10 +25,11 @@
   .table>tbody>tr>th {
     border-top: 0px;
   }
+
   .replace-options {
     margin: 0.5em 0.5em 1.5em 1em;
+    &.more-bottom-margin {
```
GarySmith: this smells like a hack, but I'm not good enough with CSS to know how to improve it. Could this instead be handled by using an appropriate CSS selector like `:last-child` or `:last-of-type` or something?

src/styles/components/serverutils.less
```diff
@@ -142,6 +142,9 @@
 .message-line {
   line-height: 2em;
   margin: 3em 0 1em 1em;
+  &.smaller-margin {
```
GarySmith: Ditto

src/utils/ModelUtils.js
```diff
@@ -66,37 +67,23 @@ export function getAllOtherServerIds (model, autoServers, manualServers, theId)
...
 
-  if(theId) {
-    let idx = model.getIn(['inputModel', 'servers']).findIndex(
-      server => server.get('id') === theId
-    );
-    //if have current id, remove current id so won't check against it
-    if (idx !== -1) {
-      modelIds.splice(idx, 1);
-    }
+  if(!isEmpty(theId)) {
```
GarySmith: Nice improvement!

src/utils/ModelUtils.js
```diff
@@ -66,37 +67,23 @@ export function getAllOtherServerIds (model, autoServers, manualServers, theId)
...
   let autoServerIds = [];
   if (autoServers && autoServers.length > 0) {
     autoServerIds = autoServers.map(server => server.id);
-    if(theId) {
-      let idx = autoServerIds.findIndex(id => id === theId);
-      // if have it in auto discovered servers, remove it.
-      if (idx !== -1) {
-        autoServerIds.splice(idx, 1);
-      }
+    if(!isEmpty(theId)) {
```
GarySmith: Ditto

GarySmith (changes requested):

src/pages/UpdateServers.js
```diff
@@ -152,26 +175,40 @@ class UpdateServers extends BaseUpdateWizardPage {
   }
 
   replaceServer = (server, theProps) =>  {
-    // update model
-    let model =
-      updateServersInModel(server, this.props.model, MODEL_SERVER_PROPS_ALL, server.id);
+    let model;
+
+    // compute host replacement has oldServer
+    let isCompute = theProps.oldServer;
```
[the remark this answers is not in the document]
guglo: yeah...it gets a little messy here. Will re-arrange some code to make a  better flow.

src/components/ReplaceServerDetails.js
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update

src/components/ReplaceServerDetails.js
guglo: will update

src/styles/components/replaceserverdetails.less
guglo: will look into it

src/styles/components/serverutils.less
guglo: will look into it

src/pages/UpdateServers.js
guglo: the props will include extra params for UpdateServer to process...think I will keep all the extra stuffs in theProps especially when most of the process logic will be moved to UpdateServer..otherwise I need to pass wipe disk, installos, user/pass and selectedserverId...and some are not necessary there.

src/styles/components/replaceserverdetails.less
guglo: tried it with :last-child which is too generic...if use last-child...it applies to all that kind of last child in the dialog which I don't want that..I am not pro CSS like Jeremy, I will ask Jeremy tomorrow.

jeremy-moffitt: If those dont work I dont think that the current implementation is all that bad, not worth spending more than 20-30 minutes on

guglo: Tried it ..since I have various sections contain server-details-container...adding last-child and last-of-type will apply multiple of last-child or last-of-type. I only need add a little bit spaces before the buttons when all the sections are expanded. So will keep the hackish way for now.

GarySmith (changes requested):
Thanks for the changes. I think it only needs the change of `isCompute` to `isComputeNode` or `hasComputeRole` for clarity

GarySmith (approved):

PR #201, opened 2018-11-01
SCRD-3497 Add PrepareController

GarySmith:
Add a new page for the prepare server workflow that is specifically for
preparing control nodes for replacement, up to the point where the OS
installation would begin.  Moved the model commit into an action so that
it will not be repeated when the page is realoaded.
Moved PrepareReplace to PrepareCompute, since the workflow for replacing
a compute node is different.

### DIFF
--- a/src/pages/ReplaceServer/PrepareController.js
+++ b/src/pages/ReplaceServer/PrepareController.js
@@ -1,0 +1,180 @@
+// (c) Copyright 2018 SUSE LLC
+/**
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+**/
+
+import React from 'react';
+import { translate } from '../../localization/localize.js';
+import { STATUS } from '../../utils/constants.js';
+import BaseUpdateWizardPage from '../BaseUpdateWizardPage.js';
+import { PlaybookProgress } from '../../components/PlaybookProcess.js';
+import { ErrorBanner } from '../../components/Messages.js';
+import { fetchJson, postJson, deleteJson } from '../../utils/RestUtils.js';
+
+class PrepareController extends BaseUpdateWizardPage {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      overallStatus: STATUS.UNKNOWN, // overall status of entire playbook and commit
+      invalidMsg: '',
+    };
+  }
+
+  setNextButtonDisabled = () => this.state.overallStatus != STATUS.COMPLETE;
+
+  updatePageStatus = (status) => {
+    this.setState({overallStatus: status});
+    if (status === STATUS.FAILED) {
+      this.setState({invalidMsg: translate('server.replace.prepare.failure')});
+    }
+  }
+
+  renderPlaybookProgress () {
+    const steps = [
+      {
+        label: translate('deploy.progress.commit'),
+        playbooks: ['commit']
+      },
+      {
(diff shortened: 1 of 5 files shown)


src/pages/ReplaceServer/PrepareController.js
```diff
@@ -0,0 +1,180 @@
...
+        name: 'ready-deployment',
+      },
+      {
+        name: 'rm-cobbler',
+        action: ((logger) => {
+          return fetchJson('/api/v1/clm/cobbler/servers')
+            .then((response) => {
+              const cobbler_server = response.find((e) =>
+                e.ip === this.props.operationProps.server.ip ||
+                e.name === this.props.operationProps.server.name);
```
guglo: this.props.operationProps.server.name should be this.props.operationProps.server.id which is the model server id matching to the cobbler name

src/pages/ReplaceServer/PrepareController.js
```diff
@@ -0,0 +1,180 @@
...
+        }),
+      },
+      {
+        name: 'config-processor-run',
+      },
+      {
+        name: 'ready-deployment',
+      },
+      {
+        name: 'rm-cobbler',
```
guglo: Just wondering ...if it is possible to make action step reusable ? Since I need to do the same.

guglo:
I also have Failed to launch playbook cobbler-deploy.yml. NOT FOUND...my ardana-service is up-to-date, also run tools/setup_env.sh

src/pages/ReplaceServer/PrepareController.js
```diff
@@ -0,0 +1,180 @@
...
+                logger('Host not present in cobbler, continuing\n');
+              }
+            });
+        }),
+      },
+      {
+        name: 'known-hosts',
+        action: ((logger) => {
+          return deleteJson('/api/v1/clm/known_hosts/' + this.props.operationProps.server.id)
+            .then((response) => {
```
guglo: I checked on my provo system, 
day0-cp1-c1-m1-mgmt ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBDSEpFr/J2HPBsUjAlfWzYwivCEGhh24I417PKCaCeg0OcdznKWBJYxp21mECJNkGgdaxb4b5Io/TAn2wVG+i28=
hostname is day0-cp1-c1-m1-mgmt.  this.props.operationProps.server.id is from servers.yml which in this case is controller1. so you can not use this.props.operationProps.server.id here But you can use server id + ip find it from  '/api/v1/clm/model/cp_internal/CloudModel.yaml'  cloudModel['internal']['servers']; in sever['hostname'] which should have day0-cp1-c1-m1-mgmt. 

 Off the topic: if you need set --limit down the road , need to use day0-cp1-c1-m1 which is in  server['ardana_ansible_host'].

src/pages/ReplaceServer/PrepareController.js
```diff
@@ -0,0 +1,180 @@
...
+        action: ((logger) => {
+          const commitMessage = {'message': 'Committed via Ardana Installer'};
+          return postJson('/api/v1/clm/model/commit', commitMessage)
+            .then((response) => {
+              logger('Model committed\n');
+            })
+            .catch((error) => {
+              const message = translate('update.commit.failure', error.toString());
+              logger(message+'\n');
+              throw new Error(message);
```
guglo: Adding commit as a step is a good idea. I might do the same.

src/pages/ReplaceServer/PrepareController.js
GarySmith: That's a good idea. I'll move some of these steps out to another class in another PR

src/pages/ReplaceServer/PrepareController.js
GarySmith: Good catch. Will fix

GarySmith:
The cobbler-deploy.yml is not yet present in the ardana-service, but that is not part of the ui change -- I'll have to make that change in the ardana-service repo independently. (Actually in another PR I will be removing the call to cobbler-deploy and calling the installui-os-providion one instead)

guglo:
@GarySmith  thanks for the changes. Since it is on going..I will push this and test the the follow-up PR .

PR #202, opened 2018-11-02
SCRD-3501 Updated to allow compute  node to be reachable

ghost:
This commit includes:

1. Updated checkPrereqs logic to allow compute node to be reachable so
it can perform instances migration.
2. If the compute node is not reachable, have a warning so users are
aware if they choose to replace the node, instances are lost.
3. Minor isCompute cleanup in RepalceServerDetails.


jeremy-moffitt (approved):
LGTM, tested modal in a local dev env

PR #203, opened 2018-11-02
 SCRD-3497 Use install os provision playbook

GarySmith:
Replace call to cobbler deploy with a call to the
install-ui-os-provision playbook, which is also used by the install
wizard.
Support defaulting and passing extra-vars in both install and update
scenarios.


GarySmith:
Note that the first two commits in this PR actually belong to PR #201 (stupid github)

PR #206, opened 2018-11-06
SCRD-3497 Add monasca, os-config steps

GarySmith:
Add additional steps to the replace server workflow for monasca
reconfiguration and osconfig playbooks.

### DIFF
--- a/src/pages/ReplaceServer/ReplaceController.js
+++ b/src/pages/ReplaceServer/ReplaceController.js
@@ -38,6 +38,13 @@ class ReplaceController extends BaseUpdateWizardPage {
 
   setNextButtonDisabled = () => this.state.overallStatus != STATUS.COMPLETE;
 
+  getHostname = (internalModel, serverId) => {
+    const matches = internalModel.internal.servers.filter(s => s.id == serverId).map(s => s.hostname);
+    if (matches.length > 0) {
+      return matches[0];
+    }
+  }
+
   componentWillMount() {
     this.setState({showLoadingMask: true});
     getInternalModel()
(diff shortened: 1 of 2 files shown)


src/pages/ReplaceServer/ReplaceController.js
```diff
@@ -190,6 +200,18 @@ class ReplaceController extends BaseUpdateWizardPage {
...
     }
 
+    playbook_steps.push(
+      {
+        label: translate('server.deploy.progress.monasca-rebuild'),
+        playbook: 'monasca-rebuild-pretasks.yml'
+      },
+      {
+        label: translate('server.deploy.progress.os-config'),
+        playbook: 'os-config.yml',
```
guglo: playbook name should be osconfig-run.yml

src/pages/ReplaceServer/ReplaceController.js
```diff
@@ -38,6 +38,13 @@ class ReplaceController extends BaseUpdateWizardPage {
 
   setNextButtonDisabled = () => this.state.overallStatus != STATUS.COMPLETE;
 
+  getHostname = (internalModel, serverId) => {
+    const matches = internalModel.internal.servers.filter(s => s.id == serverId).map(s => s.hostname);
```
guglo: Based on my experience of adding compute,   if you need set --limit,  use day0-cp1-c1-m1 which is in server['ardana_ansible_host']. server['hostname'] gives  something like day0-cp1-c1-m1-mgmt. I never run osconfig-run.yml ...you might want to run it to see what it takes for limit.

GarySmith: The operations guide says it is supposed to be server['hostname'], the one with the -mgmt extension.

src/pages/ReplaceServer/ReplaceController.js
GarySmith: Thanks, good catch

src/pages/ReplaceServer/ReplaceController.js
guglo: if you checked, that sounds good. It is kind of messy that ansible playbook limit sometimes takes hostname with -mgmt and sometimes without.

PR #208, opened 2018-11-07
SCRD-5562 remove componentWillReceiveProps hook

eapodaca:
Removes deprecated componentWillReceiveProps life cycle callback.

### DIFF
--- a/src/components/ValidatingInput.js
+++ b/src/components/ValidatingInput.js
@@ -22,7 +22,18 @@ export class ValidatingInput extends Component {
     this.state = {
       errorMsg: '',
       inputValue: props.inputValue,
+      initialValue: props.inputValue,
       showMask: true
+    };
+  }
+
+  static getDerivedStateFromProps(props, state) {
+    const takeFromProps = props.inputValue != state.initialValue && (
+      state.inputValue === props.inputValue
+    );
+    return {
+      ...state,
+      inputValue: takeFromProps ? props.inputValue : state.inputValue
     };
   }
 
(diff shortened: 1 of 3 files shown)


jeremy-moffitt:
@guglo @GarySmith  do either of you remember what the firstValidate logic was added for to begin with? I'm trying to find a scenario where it actually triggers... I'm not seeing validation until a value is added/changed

GarySmith:
@jeremy-moffitt It looks like `firstValidate` was added in this PR

jeremy-moffitt:
@GarySmith I should've phrased that differently, I'm trying to understand why we need the logic that was in the componentwillmount in the first place... though I ran into some other issues anyway

eapodaca:
I think that the componentWillReceiveProps was used because for whatever reason the yaml and j2 editors update their props late not on instantiation like this PR requires right now.

GarySmith:
Using componentWIllReceiveProps is a hack to support an anti-pattern. Here's the gist of it. If you have a component that you always instantiate in the `render` function of the parent, but use some alternate mechanism like a `show` property to selectively hide it when you don't want it shown, then the component will only be instantiated once ever.  The only kind of components that should have such a hide/show component are those with content that never changes, like maybe our loading mask. If a component needs to render differently based on properties that are passed in (i.e. nearly every react component) and it uses this crummy hide/show flag, then you have to use componentWIllReceiveProps to grab those new properties and forceably re-render. 
That is exactly the opposite of how react is designed to work.

jeremy-moffitt:
The yml files work fine, but I'm still seeing issues with the .j2 editor on the validation page, the j2 files are coming up empty for me, are you seeing differently?

GarySmith:
LGTM still, and this time I actually tried the j2 editor

PR #209, opened 2018-11-07
SCRD-5564 Remove references to `this.props` in component constructors

eapodaca:
* Remove references to `this.props` from  component contructors
* Remove unnecessary constructors

### DIFF
--- a/src/pages/ServerRoleSummary/UpdateNetworks.js
+++ b/src/pages/ServerRoleSummary/UpdateNetworks.js
@@ -29,7 +29,7 @@ import HelpText from '../../components/HelpText.js';
 class UpdateNetworks extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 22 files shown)


src/pages/ServerRoleSummary/UpdateNetworks.js
```diff
@@ -187,8 +188,9 @@ class UpdateNetworks extends Component {
     });
   }
 
-  getNetworkGroups = () => {
-    return this.props.model.getIn(['inputModel','network-groups']).map(e => e.get('name'))
+  getNetworkGroups = (props) => {
+    const model = props ? props.model : this.props.model;
```
GarySmith: getNetworkGroups is only called in one place, where props is being supplied. I don't think we should be falling back to looking at this.props.model.

PR #210, opened 2018-11-09
SCRD-3497 Handle swift ring builder manual steps

GarySmith:
If the server being replaced is the swift ring builder, there are some
steps that are currently required to be performed manually part of the
way through the process.  Added logic to determine this situation.
Added promise-handling logic to be able to interrupt the flow of the
replacement while waiting for the user to complete the manual steps.


GarySmith:
Going to update with a related change for the deploy playbook, so will create a new PR then.

PR #211, opened 2018-11-09
SCRD-3497 Handle swift ring builder, ardana-deploy

GarySmith:
If the server being replaced is the swift ring builder, there are some
steps that are currently required to be performed manually part of the
way through the process.  Added logic to determine this situation.
Added promise-handling logic to be able to interrupt the flow of the
replacement while waiting for the user to complete the manual steps.
Invoke the ardana-deploy playbook with the appropriate host limit,
depending on whether the node is a swift ring builder.


GarySmith:
I share your concern about the hosts being passed. I was using what the operations guide suggested, but until I can test it live, I have no way of knowing which is the correct one.

PR #213, opened 2018-11-09
SCRD-5569 Refactor validators

eapodaca:
Refactor validators that use props to make decisions.


GarySmith:
Thanks for making all of those changes. Looks great.

PR #216, opened 2018-11-12
Add precommit hook

eapodaca:
Add a pre-commit hook to make lint issues more surfaceable.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+# direct output to stderr
+exec 1>&2
+
+files_changed=$(git diff --cached --name-status | awk '$1 != "D" { print $2 }' | grep \.js$)
+[[ $files_changed ]] && node_modules/.bin/eslint $files_changed
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "lints": "eslint .",
     "protractor-setup": "webdriver-manager update",
     "protractor": "protractor protractor.conf.js",
+    "postinstall": "yes n | cp -i pre-commit.sh .git/hooks/pre-commit",
     "start": " webpack-dev-server --mode=development --port 2209 --hot --host 0.0.0.0",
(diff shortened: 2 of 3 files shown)


GarySmith:
I like this approach much better. Unfortunately this script doesn't seem to flag the obvious lint error that I introduced while committing. Debugging this now...

pre-commit.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# direct output to stderr
+exec 1>&2
+
+files_changed=$(git diff --cached --name-status | awk '$1 != "D" { print $2 }' | grep \.js$)
+
+echo "pre-comit > eslint"
```
GarySmith: pre-commit is messpelled here

pre-commit.sh
```diff
@@ -0,0 +1,8 @@
...
+[ ! -z "$files_changed" ] && node_modules/.bin/eslint $files_changed || echo "skipping"
```
GarySmith: `! -z` is a double-negative. It would be more direct to say `[ -n "$files_changed" ]` or even just `[ "$files_changed" ]`, or if you use `/bin/bash` in the she-bang line, then even simpler: `[[ $files_changed ]]` ... (no quotes needed when using double brackets)

GarySmith: Hmm, I don't have an `eslint` program in `node_modules/bin`. Is there a package needed in package.json for this?

GarySmith: Never mind. I missed the `.` in `.bin`. `eslint` is indeed there

GarySmith:
OK, I figured out the environmental thing I had that was causing it not to run, and it was specific to my environment, so that can be ignored. Would be nice to fix the typo since we see that *every* time we run a git commit.

GarySmith:
I know it's a stylistic thing, but you really don't need the echo statements before it is run and when it decides not to (the 'skipping' thing). Those feel like debugging rather than something that is really needed to be output every time.

PR #220, opened 2018-11-13
SCRD-5561 Update babel version

eapodaca:
* Update babel to enable fragment shorthand
* Enable function bind shorthand
* Updates webpack and loaders
* Adds JSX control statements
* Adds optional chaining
* Ads `yarn.lock` file


jeremy-moffitt:
not a review (yet) but a note that IBS will need to be updated once this lands since it changes the dependencies

PR #221, opened 2018-11-13
Fix issue with pre-commit hook

eapodaca:
* Fix issue that prevents committing files that are not js

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -3,4 +3,7 @@
 exec 1>&2
 
 files_changed=$(git diff --cached --name-status | awk '$1 != "D" { print $2 }' | grep \.js$)
-[[ $files_changed ]] && node_modules/.bin/eslint $files_changed
+if [[ $files_changed ]]; then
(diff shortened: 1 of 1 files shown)


pre-commit.sh
```diff
@@ -3,4 +3,7 @@
 exec 1>&2
 
 files_changed=$(git diff --cached --name-status | awk '$1 != "D" { print $2 }' | grep \.js$)
-[[ $files_changed ]] && node_modules/.bin/eslint $files_changed
+if [[ $files_changed ]]; then
+  node_modules/.bin/eslint $files_changed
+  exit $?
```
GarySmith: I'm surprised that this has any effect -- I thought that the original script also would have returned the error code from eslint.

GarySmith: In fact, I just experimented and found that both *should* be the same. Are you sure this works?

GarySmith: Interesting. I tried it both before and after this change and confirmed that the old version had the problem and the new one fixes it.

PR #222, opened 2018-11-14
SCRD-5570 Move password fields to forms

eapodaca:
* Move password fields to forms
* Refactor touched classes away from class property syntax
* Use function bind syntax when needed

### DIFF
--- a/src/components/Modals.js
+++ b/src/components/Modals.js
@@ -14,17 +14,15 @@
 **/
 import React, { Component } from 'react';
 import { Modal } from 'react-bootstrap';
-import { ActionButton, SubmitButton } from '../components/Buttons.js';
+import { ActionButton } from '../components/Buttons.js';
 import { translate } from '../localization/localize.js';
-import { InputLine } from '../components/InputLine.js';
-import { fetchJson, postJson } from '../utils/RestUtils.js';
-import { getAuthToken, setAuthToken, clearAuthToken } from '../utils/Auth.js';
+import { postJson } from '../utils/RestUtils.js';
 import { ErrorMessage } from '../components/Messages.js';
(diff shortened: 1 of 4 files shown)


src/components/Modals.js
```diff
@@ -110,14 +109,14 @@ class GetSshPassphraseModal extends Component {
...
-  isValidPassphrase = () => {
+  isValidPassphrase() {
     if (this.state.passphrase) {
       return true;
     }
     return false;
   }
 
-  renderErrorMessage = () => {
+  renderErrorMessag() {
```
GarySmith: This incorrectly renamed the function (missing trailing 'e')

eapodaca: Good catch.

GarySmith:
With this code in place, I am seeing errors in the Javascript console on at least one screen:
Use the ?start=installer query parameter to go into the installer, and go next until you get the "Assign Servers and Assign Server Roles" page, then click on the edit icon on one of the servers on the right-hand side of the page. When that dialog is drawn, you get errors about missing "key" props and about missing 'onAction' event handler.

eapodaca:
I've resolved the issue the `onAction` error. having trouble understanding how any of these changes would cause the `key` error.

GarySmith:
https://reactjs.org/docs/lists-and-keys.html may help if you haven't already seen it.

GarySmith:
I don't see the warnings any more

jeremy-moffitt:
trying to test this but I'm just getting the "Loading... Please Wait" whether its in Day 0 or Day 2, with no errors in the console... might be an env issue, still debugging

jeremy-moffitt:
switched to master and seeing the same thing, so I must have an env issue... will debug that and try again

PR #223, opened 2018-11-14
Replace ternary operators with If/Choose

eapodaca:
Resolves a lot of linter warnings


jeremy-moffitt:
tested the changes and they seemed fine, reminder to include a bugzilla or jira ticket in commit messages... too many without and the build team starts to reject package updates

PR #224, opened 2018-11-14
SCRD-5562 Stop tracking inputValue in internal state

GarySmith:
Instead of tracking the inputValue in both the internal state as well as
the props, rely on the parent to update the props whenever the user
changes the value, which is done via the inputAction callback.  This
matches the more standard reactJS model and fixes spurious problems
seen when updating the control programatically.

### DIFF
--- a/src/components/ValidatingInput.js
+++ b/src/components/ValidatingInput.js
@@ -21,25 +21,13 @@ export class ValidatingInput extends Component {
     super(props);
     this.state = {
       errorMsg: '',
-      inputValue: props.inputValue,
-      initialValue: props.inputValue,
       showMask: true
-    };
-  }
-
-  static getDerivedStateFromProps(props, state) {
-    const takeFromProps = props.inputValue != state.initialValue && (
-      state.inputValue === props.inputValue
-    );
-    return {
-      ...state,
-      inputValue: takeFromProps ? props.inputValue : state.inputValue
     };
   }
 
   componentDidMount() {
     if(this.props.updateFormValidity) {
-      let isValid = this.validateInput(this.state.inputValue, this.props);
+      let isValid = this.validateInput(this.props.inputValue, this.props);
       //callback function from parent to initially check
       //all inputs
       this.props.updateFormValidity(this.props, isValid);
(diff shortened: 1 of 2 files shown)


guglo:
found one issue ...not sure if it is already in master...connection info dialog won't take the inputs typing normally.

GarySmith:
In the process of adjusting the connectionInfo dialog...

GarySmith:
Addressed the reported issue. I was able to test it in most scenarios EXCEPT that succesful scenario against a oneview and a susemanager system.

guglo:
Thanks for updating the code to use better object. The validation and able and disable buttons function correctly. However I see  2 issues in connection dialog
1. It doesn't issue the discovery. I think because the sessionKey at wrong place
the saved data is like
{
	"sm": {
		"sessionKey": "",
		"secured": false,
		"checked": true,
		"apiUrl": "https://192.168.200.133:443/rpc/api",
		"creds": {
			"sessionKey": -- "123x58a5357b035abb211e620b669335319f1b99607958c6ecc60afb1260dac136d4", -- move out
			"checked": true, --remove
			"port": 443,
			"secured": false, -- remove
			"username": "admin",
			"host": "192.168.200.133"
		}
	},
	"ov": {
		"sessionKey": "",
		"secured": false,
		"checked": false,
		"apiUrl": "https://10.84.2.25",
		"creds": {
			"host": "10.84.2.25",
			"sessionKey": "LTc0ODg3OTEwMTA2ee9Nee4RtYNSF0sTdZjkI4uh79Ku7eEw", -- move out
			"username": "Administrator",
			"checked": true,  -- remove
			"secured": false --remove
		}
	}
}

The existing saved data is like the followings: 
{
	"sm": {
		"sessionKey": "121xfead09d4b32b75d63057c99c0de01e654d562098312426b2920031799d8190f4",
		"creds": {
			"secured": true,--> this is not used
			"host": "192.168.200.133",
			"port": 443,
			"username": "admin"
		},
		"checked": true,
		"apiUrl": "https://192.168.200.133:443/rpc/api",
		"secured": false
	},
	"ov": {
		"sessionKey": "LTIyMjY1OTk0NTg00dSg2pnqdSSdB2Q_UTe38NUvwr0DifH0",
		"creds": {
			"secured": true,-- this not used
			"host": "10.84.2.25",
			"username": "Administrator"
		},
		"checked": false,
		"apiUrl": "https://10.84.2.25",
		"secured": false
	}
}

2. also the dialog doesn't pick up the saved info...once user did the discovery ...and open the dialog again, he should see at least host, port, user and also checked or secured based last discovery.

PR #225, opened 2018-11-14
SCRD-5569 Support immutable lists in excludes validator

GarySmith:
Restore support of immutable lists to the createExcludesValidator
function, which will reject as invalid any input that matches a list of
values.

### DIFF
--- a/src/utils/InputValidators.js
+++ b/src/utils/InputValidators.js
@@ -185,12 +185,28 @@ export function IpInNetmaskValidator(netmask) {
 // be used in the situation where there is a fixed set of valid inputs.
 export function createExcludesValidator(values, errorMsg) {
 
+  console.assert(values !== undefined,                  // eslint-disable-line no-console
+    'Error: createExcludesValidator called without values');
+
   function validator(value) {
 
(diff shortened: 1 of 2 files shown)


src/utils/InputValidators.js
```diff
@@ -189,7 +190,7 @@ export function createExcludesValidator(values, errorMsg) {
     let exists;
     if (typeof(values) === 'object' && values instanceof Set) {
       exists = values.has(value);
-    } else if (Array.isArray(values)) {
+    } else if (Array.isArray(values) || List.isList(values)) {
```
eapodaca: what is we simplify this whole thing more?
```es6
return (values?.has(value) || values?.includes(value)) && (errorMsg || translate('duplicate.error', value)); 
```

GarySmith: I like that idea, and it does the right thing with sets and arrays, but it does not work correctly with the immutable stuff. All immutable collections, including Maps, Lists, and Sets have both a 'has' and an 'include', and using the right one depends on which of these classes are being used. I'll adjust this to handle those special cases and then default to your suggestion.

PR #227, opened 2018-11-15
Remove jquery

eapodaca:
Reduces production bundle by 600k.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "history": "^4.7.2",
     "html-webpack-plugin": "^3.2.0",
     "immutable": "^3.8.1",
-    "jquery": "^3.3.1",
     "js-yaml": "^3.10.0",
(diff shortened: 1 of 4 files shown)


GarySmith:
Great idea.  I know it wasn't part of this change originally, but we should not be checking in yarn.lock into git, since it is a generated file.

eapodaca:
Per yarn [documentation](https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control):
> All `yarn.lock` files should be checked into source control (e.g. git or mercurial). This allows Yarn to install the same exact dependency tree across all machines, whether it be your coworker’s laptop or a CI server.

GarySmith:
Let's merge this, and we can discuss separately the yarn.lock file since its use was not introduced in this change.

PR #228, opened 2018-11-16
SCRD-5901 Add package-lock.json

eapodaca:
So that users of both `yarn` and `npm` will have the same experience, we should keep the `package-lock.json` current with the yarn.lock file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ of the ardana-installer-server:
 
 ## Note!
 
-You will need the ardana-service to provide data model information.  The source 
+You will need the ardana-service to provide data model information.  The source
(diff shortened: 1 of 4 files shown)


GarySmith:
So now we will have two different generated files from two different build tools, both generated from the third file, package.json. This feels like it's going to be problematic to keep all 3 in sync.

eapodaca:
This can mostly be done automatically if you are a yarn user, added postinstall script to keep `package-lock.json` up to date.

jeremy-moffitt:
I understand the idea here, but I wonder if we're making a bit of a dependency mess to solve a problem we havent run into so far...

PR #229, opened 2018-11-16
SCRD-5073 - update to newer version of react-bootstrap

jeremy-moffitt:
Will need to update the node_modules tarball after this change as well


jeremy-moffitt:
lets wait till monday to merge this, I'll do the tarball update then to pickup any other package adjustments done recently

PR #234, opened 2018-11-19
SCRD-3480 display server status from monasca in servers page

jeremy-moffitt:
Displays monasca status up/down/unknown if Monasca is installed, shows blank column if Monasca is not present


src/pages/UpdateServers.js
```diff
@@ -93,8 +102,62 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+  checkMonasca() {
+    if(this.state.monasca === undefined) {
+      fetchJson('/api/v1/clm/monasca/is_installed')
+        .then(responseData => {
+          if(responseData.installed) {
+            getInternalModel()
+              .then((yml) => {
+                this.setState({internalModel: yml, monasca: true}, () => this.getServerStatuses());
+              }).catch((error) => {
+                console.log('error getting internal model:' + error);// eslint-disable-line no-console
```
GarySmith: If this line is executed, then the monasca state variable will remain undefined, despite the fact that we know that it is installed. That smells wrong.

src/pages/UpdateServers.js
```diff
@@ -93,8 +102,62 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+
+  getServerStatuses() {
+    if(this.state.monasca && this.state.internalModel !== undefined
+        && this.props.model !== undefined) {
+      // get the list of all servers, then load their statuses
+      // possible future enhancement: batching this
+      const serverIds = this.props.model.getIn(['inputModel','servers'])
+        .map(server => server.get('id'));
+
+      serverIds.forEach((server_id) => {
```
GarySmith: This is a potentially scalability problem. At a minimum, we might want to throttle this (with `lodash.throttle`) to avoid having too many parallel back end calls at once.

jeremy-moffitt: it turns out that throttle discards extra calls made during the throttle period, so I'm looking at alternative approaches... with throttle, depending on the settings I use, I get either 1 or 2 calls , while the other server requests get dropped

GarySmith: There is also a promise-throttle npm library to consider. If none of these pan out via a brief investiagation, we can probably defer the scalability concern to another story.

src/pages/UpdateServers.js
```diff
@@ -93,6 +102,64 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+    if(this.state.monasca && this.state.internalModel !== undefined
+        && this.props.model !== undefined) {
+      // get the list of all servers, then load their statuses
+      // possible future enhancement: batching this
+      const serverIds = this.props.model.getIn(['inputModel','servers'])
+        .map(server => server.get('id'));
+
+      serverIds.forEach(async (server_id) => {
+        let server = this.state.internalModel.internal.servers.find(s => s.id == server_id);
+        await fetchJson('/api/v1/clm/monasca/server_status/' + server.hostname)
```
eapodaca: using await here isn't really beneficial at all. I think what you want is:
```js
let responceData = await fetchJson(`/api/v1/clm/monasca/server_status/${server.hostname}`);
```
Then you can get rid of the `then` and catch calls

GarySmith:
LGTM (did not test it)

PR #237, opened 2018-11-26
SCRD-3497 Avoid passing list in extra vars

GarySmith:
When calling the install playbook, the extra-vars should contain a node
list.  The node list should always be a simple string, rather than a
python list.  Note that in the event that multiple nodes are to be
supplied (not relevant in this change), the string should be a
comma-separated list of nodes.


GarySmith:
Note that this problem was discovered when testing on our environment in the provo lab.

PR #240, opened 2018-11-28
SCRD-5561 - refactor connection dialog to use ConfirmModal

ghost:
This commit includes:
1. refactor connection dialog to use ConfirmModal so it has the
consistent look and feel as other dialogs.
2. updated to remove the react warnings regarding change uncontrolled
input to be controlled input by init with empty strings.


jeremy-moffitt (approved):
LGTM, tried it out and worked fine

PR #241, opened 2018-11-29
SCRD-6124 Call ardana-service directly

GarySmith:
Make service calls directly to the ardana-service rather than the shim.
Update URLs from v1 to v2 where necessary.


GarySmith:
I'd actually prefer not to introduce that at this point.
If I were to replace `/api/v2/` with a reference to a variable like `prefix+` or `api+`, it would only save one or two characters. I suppose another way might be to change `fetchJson` (and its friends `putJson`, etc), to always prepend this prefix, but that then locks us into *always* using `/api/v2`.
But either way we would lose something that is valuable: the current version is immediately readable without any guesswork, and it permits the ability to, say, grep for `/api/v2/playbooks` in both the UI source and the python code and find all of the places where that particular API is defined and called. To me, losing that is not worth the one of two character savings.

guglo:
newly merged pr doesn't have the new url. I will merge this and do another round of updates.

PR #243, opened 2018-11-30
SCRD-3479 Add activate/deactivate to server management for compute hosts

eapodaca:
There is still an issue with this. If the use is quick enough it is possible that the context menu will have no activate or deactivate option because the status is still loading from the backend.


eapodaca:
This can be tested against dogwood

src/pages/UpdateServers.js
```diff
@@ -95,6 +116,33 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+        return {
+          ...s,
+          internal: internalServer,
+          hostname: internalServer.hostname
+        };
+      });
+
+    let serversStatus = servers.map(s => fetchJson(`/api/v1/clm/compute/services/${s.hostname}`));
+    Promise.all(serversStatus).then((values) => {
+      let serverStatuses = forEachWithObject(values, (status, index, array, obj) => {
```
jeremy-moffitt: would a for..in loop work here rather than adding a new utility function?

PR #244, opened 2018-11-30
SCRD-6126 Serve run_dist with ardana-service

GarySmith:
When using run_dist.sh to create a locally running version of the
distributed UI, use the ardana service insted of the ardana installer
server.


GarySmith:
Note that this pulls the ardana service from its mirror on github, which has the same visibility as this repo, rather than the gerrit instance, which is not externally visible.

PR #250, opened 2018-12-06
SCRD-3480 - fix issue with false positive for monasca install

jeremy-moffitt:
Since javascript interprets a String of "false" as `true` from a boolean perspective, the Monasca install check was always being interpreted as true, which causes unworkable queries in environments where no Monasca is actually installed


jeremy-moffitt:
closing in favor of backend fix

PR #255, opened 2018-12-10
SCRD-5901 Fix some issues when fetching the monasca server status

eapodaca:
* handle missing server in internal model, not sure how that happened to me
* Handle when state cannot be translated -> 'Unknown'
* Handle exception is thrown -> 'Unknown'

### DIFF
--- a/src/pages/UpdateServers.js
+++ b/src/pages/UpdateServers.js
@@ -155,18 +155,33 @@ class UpdateServers extends BaseUpdateWizardPage {
     let internalModelServers = this.state.internalModel.getIn(['internal', 'servers']).toJS();
     for (const server_id of serverIds.values()) {
       let server = internalModelServers.find(s => s.id == server_id);
+      if (!server) continue;
       try {
         const responseData = await fetchJson('/api/v2/monasca/server_status/' + server.hostname);
-        this.setState(prevState => {
-          let serverMonascaStatuses = prevState.serverMonascaStatuses;
-          serverMonascaStatuses[server_id] = translate('server.details.status.' + responseData.status);
-          return { 'serverMonascaStatuses' : serverMonascaStatuses };
-        });
+        this.setMonascaStatus(server_id, responseData.status);
       } catch(error) {
+        this.setMonascaStatus(server_id);
         console.log('error getting server status for:' + server.hostname + // eslint-disable-line no-console
           ' -- error is:' + error);
       }
     }
+  }
+
+  /**
+   * Update the translated status of a server status
+   * @param {String} server_id The id of the server
+   * @param {String} status The current status of the server
+   */
+  setMonascaStatus(server_id, status) {
+    this.setState(prevState => {
+      let { serverMonascaStatuses } = prevState,
+        translationKey = `server.details.status.${status}`;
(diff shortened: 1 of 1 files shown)


src/pages/UpdateServers.js
```diff
@@ -155,14 +155,23 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+      if (!server) continue;
       try {
         const responseData = await fetchJson('/api/v2/monasca/server_status/' + server.hostname);
-        this.setState(prevState => {
-          let serverMonascaStatuses = prevState.serverMonascaStatuses;
-          serverMonascaStatuses[server_id] = translate('server.details.status.' + responseData.status);
-          return { 'serverMonascaStatuses' : serverMonascaStatuses };
+        this.setState({
+          serverMonascaStatuses: {
+            ...this.state.serverMonascaStatuses,
```
GarySmith: I don't understand the point of this line. Why is it trying to set the state to the same value? And in general, referring to `this.state` with a `setState` call is a race condition, since the `setState` is executed asynchronously.

src/pages/UpdateServers.js
```diff
@@ -155,14 +155,23 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+            [server_id]: (translate(`server.details.status.${responseData.status}`)
+              || translate('server.details.status.unknown'))
+          }
         });
       } catch(error) {
+        this.setState({
+          serverMonascaStatuses: {
+            ...this.state.serverMonascaStatuses,
```
GarySmith: Ditto above

src/pages/UpdateServers.js
eapodaca: This is setting the state variable `serverMonascaStatuses` to a new object with all the same properties as before (by copying them from `this.state.serverMonascaStatuses`) but with a new value for the one property with id value of `server_id` variable. This does not cause a "race condition", not possible in the browser we have only one thread of execution. Though something that we should be worried about is updating the state during a render, which can causes infinite re-renders.

GarySmith: Upon reflect, my point #1 does not apply here because you are changing a value within an object within state, which `setState` will not detect, so that part is not a problem. You just need to pay attention to #2 by making use of `prevState`

jeremy-moffitt:
latest version looks ok to me at a glance, I'm rebuilding dogwood in hopes of testing it just in case

src/pages/UpdateServers.js
```diff
@@ -155,20 +155,32 @@ class UpdateServers extends BaseUpdateWizardPage {
...
         console.log('error getting server status for:' + server.hostname + // eslint-disable-line no-console
           ' -- error is:' + error);
       }
     }
   }
 
+  /**
+   * Update the translated status of a server state
+   * @param {String} server_id The id of the server
+   * @param {String} state The current state of the server
```
GarySmith: sorry to nitpick, but this variable should be called `status` rather than `state`, to avoid confusion with the `state` you are setting. I had to read this a couple of times to figure out that it was not related to `this.state`

eapodaca:
@GarySmith see updates.

PR #256, opened 2018-12-11
SCRD-4770 Unhide server groups and nic mappings buttons when edit server in day2 UI

ghost:
This commit includes:
1. Unhide buttons related to server groups and nic mappings in
   edit server details in day2 UI.
In day2 UI:
2. The server groups used in deployed servers can not be deleted.
3. The existing networks and child server groups members in the
   server group used by deployed servers can not be deleted.
4. User can add networks or child server groups in the server group
   used by deployed servers.
5. User can not edit the server group name if the server group used
   by deployed servers.
6. The nic mappings used by deployed servers can not be deleted.
7. The existing logical names and ports in the nic mapping used by
   deployed servers can not edited or deleted.
8. The nic mapping name can not be edited if it is used by deployed
   servers.
9. User can add new logical name + ports for a nic mapping used by
   deployed servers..
10. User can still can add new server group or nic mapping
11. User can still edit or delete server group or nic mapping that is
   not used by deployed servers.

In day0 UI:
User should be able add/delete/edit server group and nic mappings.

Note: the UI goes with change in https://gerrit.suse.provo.cloud/#/c/5356/

### DIFF
--- a/src/pages/ServerRoleSummary/NicMappingTab.js
+++ b/src/pages/ServerRoleSummary/NicMappingTab.js
@@ -13,6 +13,7 @@
 * limitations under the License.
 **/
 import React, { Component } from 'react';
+import { isEmpty } from 'lodash';
 import { translate } from '../../localization/localize.js';
 import { alphabetically } from '../../utils/Sort.js';
 import { MODE } from '../../utils/constants.js';
@@ -22,9 +23,9 @@ import { List, Map } from 'immutable';
 import { NetworkInterfaceValidator, PCIAddressValidator } from '../../utils/InputValidators.js';
 import { YesNoModal } from '../../components/Modals.js';
 import { Tooltip, OverlayTrigger } from 'react-bootstrap';
-import { fetchJson } from '../../utils/RestUtils.js';
 import { ErrorMessage } from '../../components/Messages.js';
 import { LoadingMask } from '../../components/LoadingMask.js';
+import { getInternalModel } from '../topology/TopologyUtils.js';
 
 class NicMappingTab extends Component {
 
@@ -52,10 +53,10 @@ class NicMappingTab extends Component {
   componentDidMount() {
     if(this.props.isUpdateMode) {
       this.setState({loading: true});
-      // fetchJson(url, init, forceLogin, noCache)
-      fetchJson('/api/v2/model/nic_mappings_in_use', undefined, true, true)
-        .then((nic_mappings) => {
-          if (nic_mappings) {
+      getInternalModel()
+        .then((internalModel) => {
+          if (internalModel) {
+            let nic_mappings = this.getNicMappingsInUse(internalModel);
             this.setState({nicMappingsInUse: nic_mappings, loading: false});
           }
         })
@@ -67,6 +68,17 @@ class NicMappingTab extends Component {
         });
     }
   }
+
+  getNicMappingsInUse = (internalModel) => {
+    let nicmappings =
+      internalModel['internal']['servers']?.filter(server => !isEmpty(server['ardana_ansible_host']))
+        .map(server => server['nic_map']);
+    // remove the duplicates
+    nicmappings =
+      [...new Set(nicmappings.map(nMapping => JSON.stringify(nMapping)))]
+        .map(mappingStr => JSON.parse(mappingStr));
+    return nicmappings;
+  };
 
   resetData = () => {
     this.setState({
@@ -236,9 +248,8 @@ class NicMappingTab extends Component {
       // if have nicMappingInUse and it has this logical-name
       // this logic-name is in use
       let isInUse =
-        nicMappingInUse ?
-          nicMappingInUse['physical-ports'].map(port => port['logical-name'])
-            .includes(row.get('logical-name')): false;
+        (nicMappingInUse && nicMappingInUse['physical-ports'].map(port => port['logical-name'])
+          .includes(row.get('logical-name')));
 
       return (
         <div key={idx} className='dropdown-plus-minus'>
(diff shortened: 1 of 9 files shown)


jeremy-moffitt (approved):
LGTM, I think we should consider always having those buttons enabled on the add server page... is there an issue with saving the model if we do that? or are they absent to save clutter in the UI?

src/pages/ServerRoleSummary/NicMappingTab.js
```diff
@@ -38,9 +41,33 @@ class NicMappingTab extends Component {
...
+      loading: false,
+      errorMsg: undefined
     };
   }
 
+  componentDidMount() {
+    if(this.props.isUpdateMode) {
+      this.setState({loading: true});
+      // fetchJson(url, init, forceLogin, noCache)
+      fetchJson('/api/v2/model/nic_mappings_in_use', undefined, true, true)
```
GarySmith: As we discussed in person, for consistency with the rest of the UI, it probably makes more sense to just use the values here from the internal model (the output of the config processor). This API only deals with the corner case where the config processor has been run but the ready deployment playbook has not been run; if in that situation we want to ignore the config processor output and instead look in the scratch dir, then we need to do that across the entirety of the day 2 UI for consistency. That would include all of the topology pages. It would probably make more sense to handle that centrally by modifying the `getInternalModel` to ignore the cp output and just pluck out the information from the scratch directory.

GarySmith (changes requested):

src/pages/ServerRoleSummary/NicMappingTab.js
```diff
@@ -203,14 +230,21 @@ class NicMappingTab extends Component {
     this.closeDetails();
   }
 
-  renderDetailRows() {
+  renderDetailRows(nicMappingInUse) {
     return this.state.detailRows.map((row, idx, arr) => {
       const lastRow = (idx === arr.size-1);
+      // if have nicMappingInUse and it has this logical-name
+      // this logic-name is in use
+      let isInUse =
```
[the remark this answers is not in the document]
guglo: @GarySmith  ok will update

src/pages/ServerRoleSummary/NicMappingTab.js
guglo: @GarySmith per conversation, will update to call getInternalModel

guglo: @GarySmith .starting working on the changes regarding calling getInternalModel...then I am thinking...getInternalModel still goes to backend to fetch the internal model and return a big object + I need to add UI process to extract the info I need..while calling get_nic_mapping_in_use returns a much smaller object which is ready to be consumed by UI.   I am wondering whether it makes sense to change to getInternalModel now.

GarySmith: @guglo Keep in mind that the UI caches the internalModel, so it will only fetch a new one the first time after the config processor is run. So actually it would use less memory and less time to make use of the information that is already cached.

guglo: @GarySmith to resolve the consistency...I can update the backend to not use ansible resources command.

GarySmith: @guglo This function belongs in the UI, which usually has the model cached already. It does not make sense to go calculate this on the back end.

guglo: @GarySmith  ok. will update.

src/pages/ServerRoleSummary/NicMappingTab.js
```diff
@@ -68,6 +69,17 @@ class NicMappingTab extends Component {
     }
   }
 
+  getNicMappingsInUse = (internalModel) => {
+    let nicmappings =
+      internalModel['internal']['servers']?.filter(server => !isEmpty(server['ardana_ansible_host']))
+        .map(server => server['nic_map']);
+    // remove the duplicates
+    nicmappings =
+      [...new Set(nicmappings.map(nMapping => JSON.stringify(nMapping)))]
```
GarySmith: I'm having difficulty envisioning why we need to `JSON.stringify` and then undo that again on the next line with `JSON.parse`. And if there is a good reason, please comment why and what this is doing.

src/pages/ServerRoleSummary/ServerGroupsTab.js
```diff
@@ -59,6 +60,25 @@ class ServerGroupsTab extends Component {
     }
   }
 
+  getServerGroupsInUse = (internalModel) => {
+    let serverGroupList =
+      internalModel['internal']['servers']?.filter(server => !isEmpty(server['ardana_ansible_host']))
+        .map(server => server['server-group-list']);
+    // flat and remove the duplicates
+    serverGroupList =[...new Set([].concat(...serverGroupList))];
```
GarySmith: Ditto. Needs a comment about serverGroupList being a list of lists or whatever it is.

GarySmith (changes requested):

src/pages/ServerRoleSummary/NicMappingTab.js
GarySmith: After discussing via rocket chat, it appears that this is trying to find unique objects in a list, and each object has many properties and is fairly deeply nested. This is converting each object to a string and then looking for unique strings. That is fundamentally flawed in JavaScript because the order of entries in an object is not specified, and two objects containing the same contents may be converted to different strings.
I don't quite know the context for what's going on here, but NIC mapping names are unique in the model, so shouldn't it be enough to just compare the name fields?

guglo: @GarySmith  thanks for the review. Will update.

src/pages/ServerRoleSummary/ServerGroupsTab.js
guglo: @GarySmith  ok will add comments

GarySmith:
Thanks for the fix. Looks much better. Note: did not test this.

GarySmith (approved):

PR #259, opened 2018-12-12
SCRD-6374 Remove translation for logging message

ghost:
The commit includes:
1. Removed translation of the logging messages in replace server
   process
2. Updated logMessage to append new line '\n' or no new line
   based on passed in  params (SCRD-6375)
3. Removed unused messages.
4. Allowed error message from action run to show up in the page banner


guglo:
Ethan has a good idea to allow translation only to English. Will try his idea instead.

jeremy-moffitt (approved):
LGTM

src/components/PlaybookProgress.js
```diff
@@ -781,11 +781,20 @@ class PlaybookProgress extends Component {
         this.updateGlobalPlaybookStatus(playbookName, playId, STATUS.FAILED);
       }
       // if failed update caller page immediately
-      this.props.updatePageStatus(STATUS.FAILED);
+      this.props.updatePageStatus(STATUS.FAILED, error);
     }
   }
 
-  logMessage = (message) => {
+  logMessage = (message, notAppendNewLine) => {
```
GarySmith: `notAppendNewLine` is confusing. Perhaps we can change the signature to the positive case and default it to `true`, like:
```
logMessage = (message, addNewlineIfMissing=True) 
```

src/components/PlaybookProgress.js
```diff
@@ -781,11 +781,20 @@ class PlaybookProgress extends Component {
...
+    // if have not specified not to append new line break
+    // then if the message does not contain a new line
+    // break append new line break
+    if (!notAppendNewLine) {
+      var hasNewLine = /\r|\n/.exec(message);
+      if (!hasNewLine) {
+        message = '\n' + message + '\n';
```
GarySmith: I don't think we should have a newline before the message, or else all of our messages will be double-spaced.

GarySmith (changes requested):

src/components/PlaybookProgress.js
guglo: @GarySmith ok will remove it

src/components/PlaybookProgress.js
guglo: @GarySmith  that is better. will update

GarySmith (approved):

PR #261, opened 2018-12-12
SCRD-5901 Add fallback font for when 'Arial' is not available

eapodaca:
When the font Arial is not available some browsers will fall back to a serif/TimesNewRoman looking font which does not look similar to Arial at all.


eapodaca:
@jeremy-moffitt Arial being a MS font this can happen quite often in firefox on linux. Seems chrome has a hack (non-stnadard?) to detect this case and perform as this change will.

PR #262, opened 2018-12-12
SCRD-3111 Use smarter caching of internal model

GarySmith:
Loading the model that is output from the config processor, a.k.a., the
internal model, was potentially quite slow (up to about 10 seconds), so
the getInternalModel function was originally created in order to avoid
even re-querying the back-end, making the UI much faster.  But that
solution incorrectly served up old results after the config processor
was run again.  So some pages were bypassing getInternalModel and
performing the fetch directly and explicitly adding 'cache-control:
no-cache' in order to forcably re-load fresh results, and other pages
were not.

Re-factor the code to always use the no-cache header when reading the
internal model.  Note that the no-cache option has a misleading name,
since it actually compares the cached response with the ETag from the
server and only re-loads if necessary.

### DIFF
--- a/src/pages/AddServers/DeployAddServers.js
+++ b/src/pages/AddServers/DeployAddServers.js
@@ -23,7 +23,7 @@ import {
   STATUS, WIPE_DISKS_PLAYBOOK, ARDANA_GEN_HOSTS_FILE_PLAYBOOK,
   SITE_PLAYBOOK, MONASCA_DEPLOY_PLAYBOOK, ARDANA_START_PLAYBOOK
 } from '../../utils/constants.js';
-import { fetchJson } from '../../utils/RestUtils.js';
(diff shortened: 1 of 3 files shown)


GarySmith:
Updated the description of the pull request to include the original Jira ticket under which this was developed. When this is squash/merged, make sure that the title that is presented in the proposed message has the new Jira item.

guglo:
@GarySmith  from what I tested...looks like,  without the  forcing noCache...it still picks up the cached version and getting into wired states we observed before. I tried add server...
I noticed another simple way to test, in day0 ....go to change to a different model, go to page and validate the changes. Then go to day2 topology -servers you will see it still has the old servers. if force reload...then it changes to the correct servers. So getInternalModel doesn't seem to pick up the new versions of cloudModel

PR #264, opened 2018-12-12
SCRD-6374 Translation for logging messages is forced to be English

ghost:
1.Force English translation of the logging messages in replace server process
2.Updated logMessage to append new line '\n' or no new line based on passed in params (SCRD-6375)
3.Allowed error message from action run to show up in the page banner


guglo:
Abandon the changes since there are some issues Jeremy found in https://github.com/ArdanaCLM/ardana-installer-ui/pull/265 which this pr depends on.

PR #239, opened 2018-11-27
SCRD-5901 Switch to parcel

eapodaca:
What works:
:heavy_check_mark: dev build
:heavy_check_mark: source maps
:heavy_check_mark: Hot module reloading
:heavy_check_mark: less compilation
:heavy_check_mark: Asset management
:heavy_check_mark: locale bundle loading

Pros:
* Fewer direct dependencies
* Fewer configuration
* Faster (26 seconds vs 50 seconds)

Cons:
* localization bundle won't be easily editable on a deployed system
* This is all automagic, we may end up in a place where some customization we need to make is difficult or impossible.

PR #265, opened 2018-12-12
SCRD-6374 add option to force localize to use one language bundle

eapodaca:
* Copies the changes to localization to from #239.
* Adds a new function to force translate a particular key to a particular language.

This can be used as follows:
```js
import { en, translateForceLang } from '../localization/localize.js';
translateForceLang('some.key', en);
translateForceLang('some.key', 'en');
translateForceLang('some.key.with.stuff', en, 'one', 'two');
```

### DIFF
--- a/src/localization/localize.js
+++ b/src/localization/localize.js
@@ -12,56 +12,55 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 **/
-import LocalizedStrings from 'react-localization';
 import React from 'react';
 
+import * as en_bundle from './bundles/en.json';
+import * as en_bundle_branding from './bundles/en_branding.json';
+import * as ja_bundle from './bundles/ja.json';
+import * as ja_bundle_branding from './bundles/ja_branding.json';
 
-var supportedLangs = ['en', 'ja'];
-var translationData, bundlename, brandingData, bundlebrandingname, allTranslatedData, catalog = {};
+export const en = 'en';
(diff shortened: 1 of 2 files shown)


guglo:
@eapodaca  thanks for the change. I tested with Japanese, English with forced /not forced with extra strings...they all worked fine.

eapodaca:
@jeremy-moffitt cross that bridge when we get there?

src/localization/localize.js
```diff
@@ -12,56 +12,54 @@
...
 
-var findLang =
+var foundLang =
   supportedLangs.find(sLang => window.navigator.language === sLang) ||
   supportedLangs.find(sLang => window.navigator.language.substring(0, 2) === sLang.substring(0, 2));
 
-if(findLang) {
-  strings.setLanguage(findLang);
-}
-else { //default
```
jeremy-moffitt: the removal of a default language means that anyone who uses a language other than japanese or english ends up with non-translated content (even if they have English or Japanese later in their languages list)

guglo:
@jeremy-moffitt  Thanks. I spotted JP translation are not in place since there are no JP translation file,  I didn't realize that there should be a fallback on english  which is nice.

jeremy-moffitt:
this can probably be closed since https://github.com/ArdanaCLM/ardana-installer-ui/pull/259  landed, though I see you went to the effort to address some of the issues

PR #269, opened 2018-12-14
SCRD-6373 Re-join websocket after network interruptions

GarySmith:
By default the socket IO client library will detect interruptions in
communications with the server and automatically retry continually.
But when connectivity is re-established, the client is not automitically
re-added to room in which it was listening for messages.  This change
moves the code to join the room into the callback function that is
triggered whenever the connection (or re-connection) succeeds.

### DIFF
--- a/src/components/PlaybookProgress.js
+++ b/src/components/PlaybookProgress.js
@@ -275,21 +275,20 @@ class PlaybookProgress extends Component {
   }
 
   monitorSocket = (playbookName, playId) => {
+
     // Note that this function is only called after a fetch has completed, and thus
     // the application config has already completed loading, so getAppConfig can
     // be safely used here
     this.socket = io(getAppConfig('shimurl'));
     this.socket.on('playbook-start', this.playbookStarted);
-    this.socket.on(
-      'playbook-stop',
-      (stepPlaybook) => { this.playbookStopped(stepPlaybook, playbookName, playId); });
-    this.socket.on(
-      'playbook-error',
-      (stepPlaybook) => { this.playbookError(stepPlaybook, playbookName, playId); });
+    this.socket.on('playbook-stop', (stepPlaybook) => {
+      this.playbookStopped(stepPlaybook, playbookName, playId);
+    });
+    this.socket.on('playbook-error', (stepPlaybook) => {
+      this.playbookError(stepPlaybook, playbookName, playId);
+    });
     this.socket.on('log', this.logMessage);
-    this.socket.on(
-      'end',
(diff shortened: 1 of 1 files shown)


GarySmith:
For clarity, I broke submitted this as two commits -- one for reformatting the code and one with the real change. 
I found a great way to test this in dev that accurately reproduces the problem:
1. In the ardana service, make sure that if you have defined `mock_replay_speed` in your local config, that you set it to a fairly low value (I used 5), so that the playbooks will actually run for a non-trivial amount of time. Then re-start the ardana service
2. Change the `shimurl` entry in your UI's `config.dev.js` to point to some unused port other than 9085. For example, `9086`.
3. Start a local socat process that listens to that port (`9086`) and forwards it to the ardana-service on `9085`, using socat: `sudo socat tcp-listen:9086,reuseaddr,fork tcp:localhost:9085`. This will run in the foreground and can be killed by just hitting CTRL-C.

Now when you launch a long-running playbook, such as `site.yml` in the day0 interface, while it is running and the log messages are flowing, kill the socat process. The ardana-service will continue to run, and the UI will go into its retry loop. Wait a little while (say, a minute), and then re-start the `socat` process, and  you should see that the logs start flowing again. 

Note that this bug was originally seen when restarting haproxy, which only takes a second or two, so restarting socat quickly is also a realistic test as well.

src/components/PlaybookProgress.js
```diff
@@ -275,22 +275,26 @@ class PlaybookProgress extends Component {
...
+    });
+    this.socket.on('playbook-error', (stepPlaybook) => {
+      this.playbookError(stepPlaybook, playbookName, playId);
+    });
     this.socket.on('log', this.logMessage);
-    this.socket.on(
-      'end',
-      () => { this.processEndMonitorPlaybook(playbookName); });
-    this.socket.emit('join', playId);
+    this.socket.on('end', this.processEndMonitorPlaybook);
```
jeremy-moffitt: I suspect the lack of a playbook name here is the causing the problem I'm seeing with the end event

jeremy-moffitt: the following fixes the problem:
```
-    this.socket.on('end', this.processEndMonitorPlaybook);
+    this.socket.on('end', () => { this.processEndMonitorPlaybook(playbookName); });
```

GarySmith: yeah, that was bad reformatting

GarySmith:
My testing of this small fix was being subverted due to an environmental problem.

PR #270, opened 2018-12-15
SCRD-5782 Allow user to edit or delete servers at unsafe mode

ghost:
This commit includes:

1. Added option to allow user check or uncheck safe mode
2. By default it is a safe mode. When at safe mode existing servers in
the model table can not be dragged, edited or deleted.
3. If uncheck safe mode, a warning informs user that the dangers
of unsafe mode. If user wants to proceed, it is in unsafemode. when
it is unsafe mode, existing servers in the modeal table can be dragged,
edited or deleted.
4. safeMode is saved into progress.json


src/localization/bundles/en.json
```diff
@@ -585,6 +586,7 @@
     "replace.server.poweroff.warning": "The controller node must be powered off before it can be replaced since the new node will re-use the same IP address.\nPower off this node and continue",
     "replace.server.nomigration.warning": "The compute host {0} with IP {1} is not reachable, the replacement process won't be able to migrate instances. \n\n Do you still want to procceed?",
     "replace.server.compute.validate.error.title": "Validation of the newly added compute server failed",
+    "server.replace.unsafe.warning": "Unchecking safe mode will allow you do delete and modify existing servers. Deleting or modifying existing servers can potentially corrupt the cloud. \n\nDo you want to use unsafe mode?",
```
jeremy-moffitt: think you mean "to delete" not "do delete"

jeremy-moffitt (changes requested):
just need to fix the do/to typo and the rest looks good

src/localization/bundles/en.json
guglo: @jeremy-moffitt ok will fix

jeremy-moffitt (approved):
worked well, thanks for fixing the text

PR #272, opened 2018-12-18
SCRD-3501 Fixed a couple issues of replacing compute after testing using QA system

ghost:
This commit includes:
Tested on a QA system with real hardwares.

1. Fixed OS install page for replacing a compute and also added an error
message when failed.
2. Set isReachable true if the old compute server is reachable.
3. Removed the forced test condition for migrating instances.

### DIFF
--- a/src/pages/ReplaceServer/InstallOS.js
+++ b/src/pages/ReplaceServer/InstallOS.js
@@ -38,6 +38,11 @@ class InstallOS extends BaseUpdateWizardPage {
 
   updatePageStatus = (status) => {
     this.setState({overallStatus: status});
+    if (status === STATUS.FAILED) {
+      this.setState({
+        processErrorBanner:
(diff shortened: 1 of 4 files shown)


src/pages/ReplaceServer/InstallOS.js
```diff
@@ -69,7 +74,7 @@ class InstallOS extends BaseUpdateWizardPage {
       name: INSTALL_PLAYBOOK,
       payload: {
         'extra-vars': {
-          'nodelist': [this.props.operationProps.server.id],
+          'nodelist': this.props.operationProps.server.id,
```
jeremy-moffitt: this is a little confusing given the variable name, but I suppose its not really any different than before

jeremy-moffitt (approved):

GarySmith (approved):

PR #274, opened 2018-12-19
SCRD-5901 Only show monasca status column when the service is available

eapodaca:


jeremy-moffitt:
superseded by https://github.com/ArdanaCLM/ardana-installer-ui/pull/276   because updating a fork is hard

PR #275, opened 2018-12-19
SCRD-3481 - add filesystem utilization metrics from Monasca 

jeremy-moffitt:
When Monasca is available on the system, load filesystem metrics for a server when a user views the server details.

The filesystem metrics are at the bottom of the details modal below the networks list


jeremy-moffitt:
This can be tested by using dogwoods ardana-service, set `'shimurl': 'http://192.168.245.2:9085'` in your config.dev.js  , the current login pass is `Pse3EdcdkN4` (will change when updated)

guglo:
new dogwood HPCeVuYJwtXq

PR #276, opened 2018-12-19
SCRD-3481 - only show Monasca column when Monasca is installed

jeremy-moffitt:
The collapsible table had a magic number that was incorrect calculating the number of columns to display. Fixed that issue and leveraged the fix to hide the Monsaca column when Monasca is not installed

### DIFF
--- a/src/components/CollapsibleTable.js
+++ b/src/components/CollapsibleTable.js
@@ -279,7 +279,17 @@ class CollapsibleTable extends Component {
     let icon = group.isExpanded ? 'expand_more' : 'expand_less';
 
     let fillerTds = [];
-    for (let i=0; i<Object.keys(group.members[0]).length - 7; i++) {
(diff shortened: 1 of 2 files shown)


jeremy-moffitt:
This will need a squad & merge when its done due to my goofy fork->branch approach here, sorry!

config.dev.js
```diff
@@ -14,7 +14,9 @@
 **/
 
 const config = {
-  'shimurl': 'http://localhost:9085',
+  'shimurl_default': 'http://localhost:9085',
+  'shimurl_dogwood': 'http://192.168.245.2:9085',
+  'shimurl': 'http://192.168.24.18:9085'
 };
 
 export { config };
```
guglo: @jeremy-moffitt  think you don't want to checkin this changes.

PR #278, opened 2019-01-10
SCRD-5782 Handle retry and cancel when add compute

ghost:
This commit includes:

1. Added retry button when failed to run playbook during adding
compute servers..
2. When user clicks retry button, it pops up a confirmation dialog
, if user clicks yes to continue, it will rerun the failed palybook.
3. When user clicks cancel button, it pops up a confirmation dialog
with some information regarding what user needs to do to redeploy
new servers.
4. Minor lints fix

Notes: use cancel button to generate failure


jeremy-moffitt (commented):
code looks ok, havent tried it out yet

jeremy-moffitt:
tried this out locally, maybe I'm trying the wrong thing...
1. when I initiate the cancel action, it dropped me back to the server list screen with no indication anything was going on... then about 10 seconds later it jumped to the playbook list... this is very confusing to the user
2. I cancelled from the first page, and when it returned me to the server list, the server I had filled in as the replacement was in the list of computes...

src/localization/bundles/en.json
```diff
@@ -698,6 +699,10 @@
...
+    "server.deploy.failure.cancel.confirm": "Please examine the errors in the log and fix the issues encountered. \nPlease check the document's trouble shooting section on how to recover and try again. \n\n Do you want to continue?",
+    "server.addserver.failure.cancel.confirm": "Please examine the errors in the log and fix the issues encountered. Please do the following steps manually on the cloud deployer before click Yes to continue... \n1. Remove the new compute server(s) from ~/openstack/my_cloud/definition/data/servers.yml. \n2. git commit -a -m 'commite message'. \n3. ~/openstack/ardana/ansible/ansible-playbook -i hosts/localhost config-processor-run.yml \n4. ~/openstack/ardana/ansible/ansible-playbook -i hosts/localhost ready-deployment.yml  \n\nOnce it goes back the Add Server page, refresh page if it is necessary. Find servers at Manual Entry tab, update the servers and drag them to the right side table to retry deployment. \n\nPlease check the document's trouble shooting section for more details. \n\n Do you want to continue?",
```
jeremy-moffitt: I'm confused... the user manually cancelled, but we're suggesting they check for errors in the log? Also we need to clean up the wording of some of this "before click Yes" is in the wrong tense (likely should be "before clicking Yes". I would remove the word 'compute' from "Remove the new compute server" , there's a typo "commite message" should be "commit message".  I dont understand what `Once it goes back the Add Server page, refresh page if it is necessary`  is trying to tell me to do. Also `Find servers at Manual Entry tab`  is confusing

jeremy-moffitt (changes requested):
The code seems to mostly work, but upon testing I found the message to confirm cancellation to be confusing...

src/localization/bundles/en.json
guglo: The cancel button is the not the cancel button that user manually cancel (in the middle). The cancel button is the the cancel button that shows up along with next button when error happens.  I will update the wordings.

jeremy-moffitt: May need some editorial review from others, but as discussed the following may work: ```Check the log for errors and if there were any issues, resolve them before proceeding. The following steps need to be performed manually on the cloud deployer before clicking Yes to continue... \n1. Remove the new compute server(s) from ~/openstack/my_cloud/definition/data/servers.yml . \n2. git commit -a -m 'commite message'. \n3. ~/openstack/ardana/ansible/ansible-playbook -i hosts/localhost config-processor-run.yml \n4. ~/openstack/ardana/ansible/ansible-playbook -i hosts/localhost ready-deployment.yml . \n\n If the Add Server page continues to display the removed
servers, refresh the page using the web browser. To retry deployment, re-add the servers to the appropriate role, and click the "Deploy" button. \n\n Do you want to continue?```

jeremy-moffitt: Also as discussed, we can add the bit about the troubleshooting guide back in once there is a troubleshooting section in the corresponding docs

guglo: @jeremy-moffitt  thanks!

jeremy-moffitt (approved):
thanks for updating the message, looks good to me

eapodaca (approved):
LGTM

PR #279, opened 2019-01-10
SCRD-6813 Fixed issue with loading mask when replace compute

ghost:
This commit includes:
Have a specific validating state resolves the issue when
validating and other processes share the same
loading mask and interfere with each other.

### DIFF
--- a/src/pages/UpdateServers.js
+++ b/src/pages/UpdateServers.js
@@ -42,9 +42,8 @@ class UpdateServers extends BaseUpdateWizardPage {
     super(props);
 
     this.state = {
-      // this loading indicator, the possible value could
-      // be undefined, '', 'Validating Changes'
-      loading: undefined,
+      loading: false,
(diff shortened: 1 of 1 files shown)


jeremy-moffitt (commented):
one question... I see this makes loading true/false always, whereas before 'undefined' was an option as well.. is there a reason 'validating' can still be expressly set to undefined?

guglo:
Before the loading is shared between regular loading and validating... it has 3 type of values: undefined (no loading), '' (loading without text) , 'Validating Changes' (loading with text)...however the call in componentDidUpdate gets triggered slightly faster than the validation....it finishes faster and sets loading undefined before validation finishes...that is why there is a wired wait time before move to next page....the solution is to just separate them...so they don't interfere with each other.

jeremy-moffitt (approved):
think the undefined and false thing is a bit weird, but seems to work ok

PR #280, opened 2019-01-23
SCRD-6903 - consolidate service packages tabs

jeremy-moffitt:
Based on feedback during the UX review, removed the "OpenStack Packages" tab and renamed the "SUSE OpenStack Cloud Packages" tab. The latter tab already included the OpenStack services in venv form (with the same version as was displayed on the OpenStack Packages tab).

Also tweaked the spacing of tab content due to having too much space at the top of most Day 2 pages


jeremy-moffitt:
Yes its really just a deletion of the other tab, but as a git log entry I was worried that would imply the information was no longer available

src/styles/common.less
```diff
@@ -304,7 +304,7 @@ p {
 .menu-tab-content {
   margin: 2em;
   .header {
-    margin: 24px 0;
+    margin: 0px 0px 24px 0px;
```
dotNomad: This gets the header closer to the tabs, rather than having a bit of a gap, but it also puts the header text and the search bar on the Packages tab on different heights creating a bit of a disconnect in the UI.

We can apply a similar change to `.header-row > .search-group`, or apply it to the full `.header-row` and ensure other headers are wrapped in a `header-row` on other pages.

Additionally we could use `margin-bottom: 24px` instead depending on how we feel about using the specific property versus margin with multiple values.

jeremy-moffitt: used the margin-bottom suggestion and applied it to the search block as well

src/utils/RouteConfig.js
```diff
@@ -58,7 +57,6 @@ export const routes = [
   { name: translate('services'), slug: '/services',
     items: [
       { name: translate('information'), slug: '/services/info', component: ServiceInfo },
-      { name: translate('packages.openstack'), slug: '/services/openstack-packages', component: OpenStackPackages },
       { name: translate('packages.ardana'), slug: '/services/openstack-ardana', component: ArdanaPackages },
```
GarySmith: For consistency with the menu name, it would be better if this slug were renamed to /services/packages

jeremy-moffitt: good suggestion! done!

GarySmith:
LGTM

PR #281, opened 2019-01-23
SCRD-6669 Adjust layout when running playbooks

GarySmith:
Adjust layout of screens when running playbooks.  Both day 0 and day 2
re-use some of the same components, but since they have different sized
headers at the top of the screen, some of the calculations have to
adjust for these differences.

### DIFF
--- a/src/styles/components/playbookprogress.less
+++ b/src/styles/components/playbookprogress.less
@@ -52,7 +52,6 @@
   .progress-body {
     display: inline-block;
     width: 100%;
-    // min-height: 40em;
   }
   .log {
     position: relative;
--- a/src/styles/components/playbookprogress.less
+++ b/src/styles/components/playbookprogress.less
@@ -13,11 +13,23 @@
 * limitations under the License.
 **/
 .playbook-progress {
+  @day0-reserved-height: 367px;
+
   ul {
+    height: @content-height;
+
+    // When contained within a main-window (i.e. day 2), adjust
+    // the reserved height due to the thinner menu bar
+    .main-window .has-progress-bar & {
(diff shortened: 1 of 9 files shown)


guglo:
@GarySmith  Tried a couple of pages...for example add server page, it has a scroll bar now. Also on the deploy pages for adding compute, the show log link is now at the very low left corner, the next button is not visible and there is a scroll bar....same issues in  replace compute playbook processes pages.  Also noticed the loading mask seems to be off on the playbook processes pages as well. 

I am not exactly sure what changed on master...it seems to have the same issue on the my master. However I tried on my retry2 branch which is based on master of 1/11/2019(right before the trip) ...it is normal. Please take a look.

guglo:
FYI my master seems to have pulled fixLayout changes...

GarySmith:
Rebased and addressed the problem Gloria saw.

GarySmith:
> @guglo I dont see any of these changes on master....

As we discussed, this was something only in your own local repo due to an unintentional merge.

src/styles/components/playbookprogress.less
```diff
@@ -40,14 +52,18 @@
   .progress-body {
     display: inline-block;
     width: 100%;
-    min-height: 40em;
+    // min-height: 40em;
```
jeremy-moffitt: maybe just remove this?

GarySmith: Done. Thanks for catching it.

guglo:
@GarySmith  Thanks for the fixes. Tried again...they look good! The only issue is when error happens, there button goes down and scroll bar show. I will still merge this and let you take a look at that later.

PR #284, opened 2019-01-25
SCRD-4240 Add passphrase dialog to login page

GarySmith:
At the login prompt add a modal dialog to prompt for the SSH passphrase
if one is needed in order to load the private SSH key.  Improve the
text to explain how it is used and how to deal with restarts while
logged in.  Remove the ability to close the dialog without entering
a valid passphrase. Remove the ssh password dialog from the ServiceInfo
page.


GarySmith:
I noticed one thing that was not introduced by this change but that we may want to address: any error message that is shown while a modal dialog box is present is kind of obscured behind the dialog box (it is dimmed and partially hidden).

PR #285, opened 2019-01-28
SCRD-6925, SCRD-6923 Update colors in Day 0/2 UI

eapodaca:
SCRD-6925:
- Update font sizes, Update login page colors and sizes
- Use eos loading icon
- Update colors used more places
- Update nav tab colors usage

SCRD-6923:
- Switch colors to use official SUSE colors
- Change icon colors to use Material Design icon colors
- Remove unnecessary CSS
- Adjust day0 reserved height
- Update a:hover to use SUSE color


src/styles/palette.less
```diff
@@ -15,30 +15,30 @@
...
 @detailcolor: #58595B;
 
-@page-background-color: #f7f7f7;
+@page-background-color: @sb-white;
 @page-font-color: #3a3e40;
 @table-header-gradient-color: #f4f4f4;
 @table-header-color: #e7e7e7;
 @box-border-color: #cccccc;
-@info-link-color: #337ab7;
+@info-link-color: @sb-cerulean-800;
```
dotNomad: Should use `@sb-cerulean-900`

src/styles/common.less
```diff
@@ -180,12 +180,15 @@ p {
...
   cursor: pointer;
-  color: @info-link-color;
+  color: @sb-gray-900;
   &.disabled {
     color: @inactivecolor;
     cursor: not-allowed;
   }
 }
+.info-banner .material-icons {
+  color: @sb-blue-800;
```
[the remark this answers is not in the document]
dotNomad: Similar change needs to be done for the `.logout-btn`

src/styles/common.less
```diff
@@ -213,13 +216,13 @@ p {
 }
 
 .login-page {
```
dotNomad: Color on the login page should be white. According to the EOS - black and colored backgrounds require white text.

src/styles/palette.less
eapodaca: This is the same link color used on the eos design app

src/styles/common.less
```diff
@@ -356,13 +362,13 @@ p {
...
 a {
-  color: #337ab7;//bootstrap 4 changed its defaul to 007bff, change it back here
+  color: @sb-cerulean-900;//bootstrap 4 changed its defaul to 007bff, change it back here
 }
 
 .alert.show:not(.alert-danger):not(.alert-success){ //in bootstrap3, this is handled via alert-info, but the alert dialog is not the same with bootstrap4
-  color: #31708f;
-  background-color: #d9edf7;
-  border-color: #bce8f1;
+  color: @sb-blue-800;
```
GarySmith: This basically is black text on a dark gray background. This cannot be right.

dotNomad: This does show as a 9.44 Contrast Ratio which is actually higher than our text color on the white background (5.77 for reference). Still a good catch, and something we should be aware we are changing.

src/styles/components/loadingmask.less
```diff
@@ -24,6 +24,9 @@
   z-index: 9000;
   background: @defaultshadow;
   opacity: 0.7;
+  .eos-icon-loading {
+    border-top-color: @sb-green-500;
```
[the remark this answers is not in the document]
eapodaca: Switched to the default colors provided by `eos-icon-loading`

jeremy-moffitt:
progress_fubar

jeremy-moffitt:
also something in this change has added a permanent scrollbar when I'm in Day-0 mode

jeremy-moffitt:
wow_thats_blue
I dont think this is right...

eapodaca:
> wow_thats_blue
> I dont think this is right...

Please be more specific.

jeremy-moffitt:
That blue is incredibly bright... I would either use what we had before, or the lightest blue in the brand set... 
@dotNomad do we have guidance about whether its ever ok to use non-brand colors intentionally?

jeremy-moffitt:
checked the Day2 pages and they look good except the Topology->Network  tab overflows off the screen to the right slightly. I'm not sure a good solution to this.
As discussed in the scrum, the scrollbars on the Day 0 content probably mean adjustments to the individual pages ahead of merging this

jeremy-moffitt:
> Hello, I can suggest one of the options, imo B is maybe too intrusive but A should do the work.
@en3sis   You don't think thats a lot of bright blue? It *really* jumps out at the user. If we put a 50% opacity over it it looks pretty good, but doesnt match what you've presented here

en3sis:
@jeremy-moffitt I agree that with 50% opacity looks less bright  but it's also true that you get that feeling because of looking at it for that long, in a normal scenario it should be different since you just check the information and move on =)
Maybe you guys can try that way, tomorrow I'll check it with the team and let you know.

en3sis:
@jeremy-moffitt @dotNomad 
I had the meeting with my team:
Since we didn't defined yet the notifications/alerts using the bootstrap colors it's fine, or if you still want to use the blue and add opacity it's also ok.

I think the main color part would be about the use of green color.

jeremy-moffitt:
the latest version of this has a scrollbar on every Day0 page for me

jeremy-moffitt:
not sure if its tied to this change, I'll have to check master, but the tooltip for "wipe data disks" on the deployment tab of "Review Configuration Files" flickers constantly unless your mouse is at the very bottom of the icon... I'll try it out on master and update comment (note that it works correctly on the demo machine which is months old)

eapodaca:
Just pushed update for the following issues:
1. Fixed the issue with password reveal on login screen.
2. changed the `material-icon` class to default to md-dark color
3. login button has `btn` class
4. `.context-menu-container.menu-item` hover now sets color to black for better contrast.

remaining issues:
1. scrollbars
2. unifying button colors with `@sb-green-500`
3. Didn't address `btn-clickable` because it does not appear to be used anywhere

dotNomad:
- Scrollbars should be fixed in #294
- Unifying button colors should be easier once we get the guidelines, which should be very soon.
- You can find `btn-clickable` under Services > Configuration and expanding one of the services. It seems like these are created in `ServiceTemplatesTab.js` where an ActionButton is given type `clickable`

jeremy-moffitt:
Day2  content looks fine with this change... the Day0 stuff has the scrollbar issues that @dotNomad indicates are resolved in #294 ... will try out that change and possibly merge them both and then tackle any issues that are still present

jeremy-moffitt:
per discussion, @dotNomad 's changes from #294 have been merged on top of this one. I'm going to do another sanity check, as is Jordan, after which we will merge this set if all looks well

jeremy-moffitt:
worth looking at in Day-0 :
1. font size on the model selection cards
2. font size on the playbook progress screen (tentative agreement to change from 1.1em to 1.2em)
3. icon placement on the info blocks (current vertically aligned 'bottom' should be 'middle')

No issues in the Day2

jeremy-moffitt:
latest changeset looks good to me... we should check with @en3sis about possibly using a different background color in the log viewer, but should not hold up this change on that discussion.
will merge once @dotNomad has had a chance to review unless he finds additional changes needed

eapodaca:
I've updated the step 3 cards to match step 2.

dotNomad:
Looks great!

PR #287, opened 2019-01-29
SCRD-6925 Writing style guide updates

eapodaca:
* Use sentance case
* Use `and` instead of `&`
* Use `:` instead of `-`


src/localization/bundles/en.json
```diff
@@ -705,9 +704,9 @@
     "server.addserver.installos.cancel.confirm": "This will cancel the OS installation for the new compute server(s). \n\nDo you want to continue? ",
 
     "table.column.header.id" : "Server ID",
-    "table.column.header.ip-addr" : "IP Address",
-    "table.column.header.server-group" : "Server Group",
-    "table.column.header.nic-mapping" : "NIC Mapping",
-    "table.column.header.mac-addr" : "MAC Address",
+    "table.column.header.ip-addr" : "IP address",
+    "table.column.header.server-group" : "Server group",
+    "table.column.header.nic-mapping" : "NIC ,apping",
```
jeremy-moffitt: ,apping ain't what it used to be

src/localization/bundles/en.json
```diff
@@ -213,42 +212,42 @@
...
+    "server.details.serialnumber": "Serial number:",
     "server.details.ilo": "ILO: ",
     "server.details.header.interface": "Interface",
-    "server.details.header.ipv4": "IP Address",
+    "server.details.header.ipv4": "IP address",
     "server.details.header.netmask": "Netmask",
     "server.details.header.broadcast": "Broadcast",
-    "server.details.header.hardware_address": "Harware Address",
-    "server.details.header.module": "Driver Module",
+    "server.details.header.hardware_address": "Harware address",
```
dotNomad: Looks like we still have the `Harware` misspelling?

PR #288, opened 2019-01-30
SCRD-5292 Added handling for encryptKey

ghost:
This commit includes the following features: 

1. Added a check if the cloud configuration is encrypted in Replace Server, Add Servers, Delete/Activate/Deactivate Compute,  Service Configuration and Service Info Page . 

2. If the cloud configuration is encrypted, when replace controller or compute server, the replace detail will have an extra required input for encrypt key; when add compute servers, the edit server detail will have an extra required input for encrypt key; when install OS, the select server page will have an extra required input for encrypt key; when update service configuration, it will have an extra required input for encrypt key.

3. If the cloud configuration is encrypted,  encryptKey has not been set, it will prompt user to input encrypt key. 

4.  If the cloud configuration is encrypted,  Replace Server, Add Servers , Delete/Activate/Deactivate Compute , isEncrypted and encryptKey are saved as global vars but not persisted. Also during playbook running process, if user refresh the process page, user will be prompted to re-enter encryptKey. 

5. If the cloud configuration is encrypted,  for Service Configuration and Service Info, during playbook running process, there is no mechanism to re-enter encryptKey when user refresh. 

6. Some minor lints, typo changes and missing method call fixes.


guglo:
Test encrypted can go with https://gerrit.suse.provo.cloud//#/c/5561
add in local.cfg to overwrite default false
[testing]
is_encrypted = true

src/localization/bundles/en.json
```diff
@@ -69,6 +69,9 @@
     "get.passphrase.invalid": "SSH passphrase provided is incorrect. Please provide the correct SSH passphrase.",
     "get.passphrase.error": "Unable to use the provided SSH passphrase: {0}.\nPlease correct the problem and try again.",
     "id.with_colon": "ID:",
+    "common.encryptkey.message": "Please provide encryption key for running playbooks",
+    "common.encryptkey.title": "Provide Encryption Key",
```
eapodaca: Please use sentence case for this.

src/pages/UpdateServers.js
```diff
@@ -386,6 +385,11 @@ class UpdateServers extends BaseUpdateWizardPage {
     // it to complete before moving on.
     await this.props.updateGlobalState('model', model);
 
+    // save the encryptKey in global cache so it could be retrieved later
```
GarySmith: I believe that the constructor needs something like:
```
  this.state = {
     ...this.state
```
in order to be consistent with the other components that extend the baseupdatewizard

src/UpdateWizard.js
```diff
@@ -62,14 +63,18 @@ class UpdateWizard extends InstallWizard {
       // operationProps record all the bits that need to run to update
       // TODO how to deal with install password
       operationProps: undefined,
+
+      encryptKey: '',
```
GarySmith: It would seem that `isEncrypted` could be derived from `encryptKey`, e.g. it is exactly the same as the expression: `not isEmpty(encryptKey)`, and thus `isEncrypted` is basically duplicated data that should be removed. 

If I am misunderstanding this, then it is worth a comment in the code explaining why two variables are needed here instead of one.

GarySmith (changes requested):

GarySmith:
Also, the commit comment could use some attention. It is hard to read, not line-wrapped correctly, has lots of repetition, and typos (like 'peristed' and 'encrpted').

src/pages/ReplaceServer/ReplaceController.js
```diff
@@ -32,11 +32,12 @@ class ReplaceController extends BaseUpdateWizardPage {
     super(props);
 
     this.state = {
+      ...this.state,
       overallStatus: STATUS.UNKNOWN, // overall status of entire playbook and commit
       invalidMsg: '',
       showLoadingMask: false,
 
-      ringBuilderConfirmation: undefined,
+      ringBuilderConfirmation: undefined
```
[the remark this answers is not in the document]
guglo: Not intentional..will restore it.

src/pages/UpdateServers.js
guglo: ok, will update.

src/UpdateWizard.js
guglo: thanks @eapodaca  for the explanation. I will update with comments.

guglo:
@GarySmith  @eapodaca  thanks for the review. I will update the commit messages and address review comments.

jeremy-moffitt:
tested without the key forced to on, will test next with it on... still need to piece through the code as well.... @guglo was there a real system with this enabled to test with too?

GarySmith (approved):

guglo:
@jeremy-moffitt  To test is_encrypted API call in development env, from ardana-service dir

jeremy-moffitt (approved):
tried it out locally, worked as advertised

PR #293, opened 2019-02-01
SCRD-6838, SCRD-4498 on complete show links

eapodaca:
Get links for complete screen from the keystone catalog.

### DIFF
--- a/src/localization/bundles/en.json
+++ b/src/localization/bundles/en.json
@@ -439,6 +439,7 @@
     "complete.message.body1": "Congratulations! You've successfully deployed your cloud.",
     "complete.message.body2": "This is the information about your cloud:",
     "complete.message.body3": "Cloud model: {0}",
+    "complete.message.body4": "Please wait a few minutes for the cloud deployment to be complete before accessing the {0}.",
     "complete.message.link.heading": "Useful links:",
     "complete.message.link1": "Openstack Horizon UI",
     "complete.message.link2": "Ops Console UI",
--- a/src/localization/bundles/en_branding.json
+++ b/src/localization/bundles/en_branding.json
@@ -4,7 +4,6 @@
     "day2.product.title": "CLM Admin Console",
     "day2.product.name": "CLM\nAdmin Console",
     "packages.ardana": "Packages",
-    "services.header.packages.ardana" : "Installed SUSE Cloud Packages",
-    "complete.message.refresh": "Please wait a few minutes for the cloud deployment to be complete before refreshing the browser in order to access the CLM Admin Console."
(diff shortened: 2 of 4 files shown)


GarySmith:
> Note that this won't work for astack deployments, but does produce the correct Horizon URL for "real" deployments. 

Can you elaborate on that a bit?

> ...figuring out why the config file is missing the values in the first place

If you are referring to the ardana service config file, the reason it does not have these values is that they are only populated in the config file by the config processor, which is run as part of the deployment page, and they are not available to the ardana-service until it is restarted or the config file is forceably re-read, at which point the ardana service would be listening on another interface that is not available to the UI, etc.

jeremy-moffitt:
> > Note that this won't work for astack deployments, but does produce the correct Horizon URL for "real" deployments.
> 
> Can you elaborate on that a bit?

on an astack system the urls show up as `https://myardana.test:443` which doesn't help the user at all

jeremy-moffitt:
based on Garys comment about the config processor, reading these from keystone and trimming the ops console URL may be the most reliable approach... so that should be a modest tweak from the current version of this change

eapodaca:
This change should be trimming the path for both horizon and the opsconsole urls. As an alternative we could have the service request for the endpoints read and parse the config file from disk.

jeremy-moffitt:
The astack fix worked, not sure I understand how the trimming is taking place (note that the Horizon url is correct coming out of keystone)

guglo:
@eapodaca  question, when keystone is up...in day0 do we need to use user/pass to access endpoint list?

eapodaca:
@jeremy-moffitt please re-review this has drastically changed since you reviewed it last.

I have updated the `/api/v2/external_urls` `ardana-service` endpoint to return the urls from the internal model.

jeremy-moffitt:
@eapodaca does this require https://gerrit.suse.provo.cloud//#/c/5590/  to have landed to function?

eapodaca:
> @eapodaca does this require https://gerrit.suse.provo.cloud//#/c/5590/ to have landed to function?

Not required but with that change the bugs are 100% fixed.

jeremy-moffitt:
tried this out with the gerrit change, and it does load the links... but it takes some time... at first I thought it was broken, but after waiting it out, the links did show up eventually.

jeremy-moffitt:
overall this looks ok to me, though its taking 10 seconds to load the links, so if that time to load cant be reduced, then we should put a loading indicator of some sort where the "useful links" are (I'm hesitant to mask off the modal as it wouldnt be very obvious why thats happening)...  open to ideas... maybe "Loading useful links..." that changes into the current "Useful links:" section.

PR #295, opened 2019-02-05
SCRD-7374 properly load data on component mount

eapodaca:
Because of props added to the page after an update wizard some items didn't get loaded properly.


eapodaca:
This code as written requires at least one expanded group be open. The last one cannot be closed.

Updated to address this and some undefined exceptions. As well as restoring the expand state after wizard flow.

dotNomad:
Worked for me on a development setup as well :+1:

PR #297, opened 2019-02-05
Merge Master to stable/pike in support of branding changes (SCRD-6925)

jeremy-moffitt:
The branding changes will require IBS to be updated with a new node_modules.tar.bz2 file reflecting the changes to package.json. In order to get the update for Devel:Cloud:8:Staging, the merge upstream (here in git) needs to have taken place. 

No new code changes, just merged from master and resolved one trivial conflict (the renderFooterButtons method in src/pages/AddServers/DeployAddServers.js was added on master and for some reason was seen as a conflict in stable/pike, though the conflicting code was empty)


dirkmueller:
Branding changes are a no-go on maintained branches. Why are we doing this?

jeremy-moffitt:
> Branding changes are a no-go on maintained branches. Why are we doing this?

1. I was unaware of that, though thinking about it I suppose it makes sense
2. It will be difficult to deliver any of the late-arriving Day2 UI functionality back into Cloud 8 separate from the branding changes due to the pervasive nature of the branding updates. 

Probably need some guidance from T.R. about whether its still desired to deliver additional Day2 work back into Cloud 8 at this timeframe... I see 2 options:
1. Stop backporting Day2 deliveries to Cloud 8, advertise what we have now as a tech preview of whats coming in Cloud 9
2. Continue to backport the changes to keep the functionality equivalent, at the cost of having some branding changes land in Cloud8

I'm fine with either. I'll email TR (who I believe is out for another week) and CC @dirkmueller

eapodaca:
@jeremy-moffitt It should be possible to use `git revert` on this branch to revert the branding commits. There were only 2 or 3 of them.

jeremy-moffitt:
@eapodaca that would likely work in the short term, but the merge conflicts going forward would likely get more and more complex... its worth evaluating at least whether there is value in continuing to backport to Cloud8 before deciding

dirkmueller:
@jeremy-moffit do not backport non bugfixes to stable/pike anymore.

PR #301, opened 2019-02-11
SCRD-3498 - remove specific doc section number reference

jeremy-moffitt:
replaced outdated number reference with section header per comments in SCRD-3498


jeremy-moffitt:
Sorry about that! Saw it as low hanging fruit from the JIRA comment, made the change then walked away from my desk for 45 minutes, came back, pushed it up, then saw the updated comments on the ticket :(

PR #304, opened 2019-02-12
Updated to deal with failed old compute (SCRD-6686)

ghost:
This commit includes:

1. When old compute is not reachable, ardana-gen-hosts-file playbook
will skip the old compute during replacement process.

2. Updated to only migrate instances when old compute is reachable
and have instances.

3. Updated to run nova-host-evacuate playbook when old compute is
not reachable, have instances and user confirmed that compute hosts
are configured with block storage or  shared storage.

4. If there are no instances, no migration or evacuation of instances
will be initiated.

5. Some minor names update.


guglo:
need to go with https://gerrit.suse.provo.cloud/#/c/5628 which is still in primitive stage.

src/pages/UpdateServers.js
```diff
@@ -752,7 +795,14 @@ class UpdateServers extends BaseUpdateWizardPage {
 
   proceedOldComputeNotReachable = () => {
     this.setState({
-      showNoMigrationWarning: false, showReplaceModal: true});
+      showComputeNotReachableWarning: false, showReplaceModal: true});
+  }
+
+  cancelOldComputeNotReachable = () => {
+    this.setState({
```
GarySmith: Changing three state variables to different values (some false, some undefined) just from cancelling one message suggests that there are too many intertwined states in this function. While I don't expect this to change now, please consider simplifying this if you have to make any substantial changes to this functionality.

src/pages/ReplaceServer/DisableComputeServiceNetwork.js
```diff
@@ -339,18 +339,23 @@ class DisableComputeServiceNetwork extends BaseUpdateWizardPage {
       playbooks: [REMOVE_FROM_AGGREGATES]
     }];
 
-    if (this.props.operationProps.oldServer.isReachable && this.props.operationProps.server) {
+    if (this.props.operationProps.oldServer.isReachable &&
```
GarySmith: Since two of the three conditions are the same between the if and else-if, it would be more straightforward to do something like:
```
if (this.props.operationProps.oldServer.hasInstances && this.props.operationProps.server) {
  if (this.props.operationProps.oldServer.isReachable) {
     steps.push(...some stuff)
  } else {
     steps.push(...that other stuff)
  }
}
```

src/pages/ReplaceServer/DisableComputeServiceNetwork.js
```diff
@@ -375,14 +380,30 @@ class DisableComputeServiceNetwork extends BaseUpdateWizardPage {
     }];
 
     // migrate instances when old compute host is reachable
-    if(this.props.operationProps.oldServer.isReachable && this.props.operationProps.server) {
+    // and has instances
```
GarySmith: Ditto above statement -- simplify these nearly-identical if-conditions.

GarySmith (approved):

src/pages/ReplaceServer/DisableComputeServiceNetwork.js
guglo: @GarySmith ok will update

src/pages/ReplaceServer/DisableComputeServiceNetwork.js
guglo: @GarySmith ok

src/pages/UpdateServers.js
guglo: @GarySmith there a 2 situation to deal with has instances , has no instances and failed to get instances. When failed to get instances , that complicates this a little bit. I will see if there is way to simplify this.

src/pages/UpdateServers.js
```diff
@@ -741,6 +770,20 @@ class UpdateServers extends BaseUpdateWizardPage {
       });
   }
 
+  checkInstances = async (hostname) => {
```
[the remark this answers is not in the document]
guglo: @GarySmith will try to simplify

src/pages/UpdateServers.js
GarySmith: I just re-read the code, and there are only two variables (not three), so it's not really that bad. I don't want you to spend a lot of time reworking this, but, maybe this can be simplified (if possible) in a later PR if time permits.

guglo: Ok good to know. I plan to come back to the UI code once I finished the playbook

guglo:
Playbook is in review https://gerrit.suse.provo.cloud//#/c/5628/
If want test not reach with instances : ardana-service/tools/compute-mock-data.json add one instance in "get_instances" like 
"get_instances": [{
    "id": "8279e65d-6e87-4a50-b789-96edd753fbb2",
    "name": "test3"
  }]

guglo:
Note: deal with not reachable in replace compute. However deactivate reuses disable compute need to address the same problem. Will deal with  deactivate in different PR.

jeremy-moffitt (approved):

PR #305, opened 2019-02-12
SCRD-6376  Updated to use playbook constants

ghost:


src/pages/CloudDeployProgress.js
```diff
@@ -15,8 +15,14 @@
...
-import { PRE_DEPLOYMENT_PLAYBOOK, DAYZERO_SITE_PLAYBOOK,
-  SITE_PLAYBOOK, STATUS } from '../utils/constants.js';
+import {
+  PRE_DEPLOYMENT_PLAYBOOK, DAYZERO_SITE_PLAYBOOK,
+  CONFIG_PROCESSOR_RUN_PLAYBOOK, READY_DEPLOYMENT_PLAYBOOK,
+  NETWORK_INTERFACE_DEPLOY_PLAYBOOK, NOVA_DEPLOY_PLAYBOOK,
+  IRONIC_DEPLOY_PLAYBOOK, MAGNUM_DEPLOY_PLAYBOOK, ARDANA_STATUS_PLAYBOOK,
+  MONASCA_AGENT_DEPLOY_PLAYBOOK, MONASCA_DEPLOY_PLAYBOOK,
+  SWIFT_DEPLOY_PLAYBOOK, MONASCA_TRANSFORM_DEPLOY_PLAYBOOK, CEPH_DEPLOY_PLAYBOOK,
+  CINDER_DEPLOY_PLAYBOOK, SITE_PLAYBOOK, STATUS } from '../utils/constants.js';
```
eapodaca: might be worth while doing a `import * as constants from '../utils/constants';` That way reference can be something like `constants.CONFIG_PROCESSOR_RUN_PLAYBOOK` and we don't have to keep this huge list here.

src/pages/ReplaceServer/ReplaceController.js
```diff
@@ -15,7 +15,14 @@
...
 import { translate } from '../../localization/localize.js';
-import { INSTALL_PLAYBOOK, STATUS } from '../../utils/constants.js';
+import {
+  CONFIG_PROCESSOR_RUN_PLAYBOOK, READY_DEPLOYMENT_PLAYBOOK,
+  ARDANA_SSH_KEYSCAN_PLAYBOOK, BM_POWER_STATUS_PLAYBOOK,
+  BM_REIMAGE_PLAYBOOK, COBBLER_DEPLOY_PLAYBOOK,
+  BM_POWER_UP_PLAYBOOK, BM_WAIT_FOR_SSH_PLAYBOOK, WIPE_DISKS_PLAYBOOK,
+  ARDANA_DEPLOY_PLAYBOOK, MONASCA_REBUILD_PRETASKS_PLAYBOOK,
+  OSCONFIG_RUN_PLAYBOOK, CEILOMETER_RECONFIGURE_PLAYBOOK,
+  INSTALL_PLAYBOOK, STATUS } from '../../utils/constants.js';
```
eapodaca: Same as mentioned above.

eapodaca (changes requested):
Other than noted, looks good.

dotNomad (changes requested):
- Seems like the Playbook constants usage always looks like `CONSTANT + '.yml'`. Should we include those in the constants themselves? We do this in a few local constants that aren't in `constant.js`
- In a handful of files we have a local constant instead of using one in `constant.js`. Should we switch those over to `constant.js` as well?

src/pages/CloudDeployProgress.js
guglo: @eapodaca  good idea ...will do

src/pages/ReplaceServer/ReplaceController.js
guglo: ok

src/pages/AddServers/PrepareAddServers.js
```diff
@@ -63,10 +65,10 @@ class PrepareAddServers extends BaseUpdateWizardPage {
       playbooks: ['commit']
```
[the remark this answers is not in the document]
guglo: @dotNomad will update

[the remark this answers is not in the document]
guglo: @dotNomad  
1. The API call need just name without '.yml' while step needs '.yml' to apply update during progress. 
2. If the name is used in multiple files...then they belong to common constants. Some action names are local I will just keep them in local. For playbooks...I move them to constants since it is too much to identify if they are used in multiple places or not.

dotNomad:
1. Yep I see that these are used with and without the `.yml`, missed that the first time.
2. Sounds good, just wanted to clarify that that was the decision.

Looks good to me :+1:

dotNomad (approved):

GarySmith (approved):

eapodaca (approved):

PR #306, opened 2019-02-14
SCRD-7653 disable update button when encrypt key is empty when configure service

ghost:
Fixed the logic to disable update button when cloud is encrypted and encrypt key is empty.


GarySmith (approved):
LGTM. Caveat: did not test it.

jeremy-moffitt (approved):
tried it out locally with and without forced encryption key, seemed to behave correctly

PR #307, opened 2019-02-14
Update button styling (SCRD-6923)

dotNomad:
Additional complexity was necessary as the guidelines used Bootstrap 3, and we are using Bootstrap 4.


jeremy-moffitt:
We may want to update the login button to have white text on green background to match the other buttons. Otherwise looks good to me.

PR #309, opened 2019-02-14
Change tooltip placement to auto (SCRD-7643)

jeremy-moffitt:
prevents the "tooltip flash" on certain pages. Does mean that tooltip placement is now dynamic and will not default to the same position for everything, but seems a reasonable tradeoff.


GarySmith:
LGTM

PR #310, opened 2019-02-14
Permit retry when replace controller fails (SCRD-6386)

GarySmith:
If the replace controller workflow fails at any point, provide the
user a retry button to resume processing at the point of failure.
Re-word the message that appears when the user wants to cancel without
retrying.


src/pages/ReplaceServer/ReplaceController.js
```diff
@@ -371,8 +371,7 @@ class ReplaceController extends BaseUpdateWizardPage {
   }
 
   render() {
-    //if error happens, cancel button shows up
-    let cancel =  this.state.overallStatus === constants.STATUS.FAILED;
+    let failed = (this.state.overallStatus === STATUS.FAILED);
```
guglo: you need to keep the consttnts. otherwise it will have undefined error and lint error as well

src/localization/bundles/en.json
```diff
@@ -712,7 +712,7 @@
     "server.deploy.progress.delete_network_agents.failure": "Error: Failed to delete network agents for compute host {0}. {1}",
     "server.deploy.installos.failure": "Failed to install SLES OS on compute host {0}.",
     "server.deploy.retry.confirm": "Retry will rerun the failed playbook. Check the log for errors and if there were any issues, resolve them before proceeding. \n\n Do you want to continue?",
-    "server.deploy.failure.cancel.confirm": "Check the log for errors and if there were any issues, resolve them before proceeding. \n\n Do you want to continue?",
+    "server.deploy.failure.cancel.confirm": "Cancelling will close the window, and any error messages will no longer be available to view.\n\nDo you want to cancel?",
```
guglo: cancel confirm doesn't apply to your screen anymore.

GarySmith: I know this does not apply any more, but the old message was incorrectly worded, making it appear the they were "proceeding" and "continuing" instead of cancelling. Going back to the summary page is certainly "closing the window", because we have a button there that says "close" under normal circumstances.

src/pages/ReplaceServer/ReplaceController.js
GarySmith: Good catch. I missed this when resolving merge conflicts.

src/localization/bundles/en.json
guglo: @GarySmith  that is fine with me.

PR #313, opened 2019-02-16
Add loading icons to deployment page (SCRD-6923)

dotNomad:


dotNomad:
Small adjustments to Day 0 based on UI/UX team's feedback.

src/styles/pages/AssignServerRoles.less
```diff
@@ -105,7 +105,6 @@
     .tab-pane {
       margin-top: 0.5em;
       max-height: 36em;
-      overflow: auto;
```
guglo: @dotNomad  can not remove this, it will overflow the pane without scroll bar.

src/components/PlaybookProgress.js
```diff
@@ -266,6 +266,12 @@ class PlaybookProgress extends Component {
       if (status === STATUS.COMPLETE) {
         return (<li key={index} className={statusClass}>{step.label}
           <i className='material-icons succeed-icon'>check_circle</i></li>);
+      } else if (status === STATUS.IN_PROGRESS) {
+        return (
+          <li key={index} className={statusClass}>
+            {step.label} <i class="eos-icons eos-icon-loading"></i>
```
guglo: change class to be className

dotNomad: Great catch!

dotNomad:
Was able to fix the hover flickering with the tooltips by applying a `popperConfig`, setting the `boundariesElement` to window instead of the default scrollParent. So the tooltips can pop up outside of the bounds of the parent element. 
[popper.js boundariesElement](https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.boundariesElement)

src/components/Buttons.js
```diff
@@ -260,74 +260,81 @@ export class ItemMenuButton extends Component {
...
+      <td className='actions'>
+        <p onClick={this.props.clickAction}>
+          <OverlayTrigger
+            placement='bottom'
+            overlay={
+              <Tooltip id='edit-row'>
+                {translate('common.edit')}
+              </Tooltip>
+            }
+          >
```
guglo: you might want to add popperConfig={preventOverflowPopperConfig} here

dotNomad: :laughing: How did I miss that? Good catch!

PR #314, opened 2019-02-19
Remove broken goback logic when token expires (SCRD-5747)

jeremy-moffitt:
The logic to "goBack" to the page that timed out isn't working and causes the user to have to login twice (only to be taken to the default landing page anyway). For now, removed the extra logic, which fixes the double login problem.

### DIFF
--- a/src/pages/Login.js
+++ b/src/pages/Login.js
@@ -16,7 +16,7 @@ import React, { Component } from 'react';
 import { translate } from '../localization/localize.js';
 import { fetchJson, postJson } from '../utils/RestUtils.js';
 import { setAuthToken, clearAuthToken } from '../utils/Auth.js';
-import { navigateTo, navigateBack, wasRedirectedToLogin } from '../utils/RouteUtils.js';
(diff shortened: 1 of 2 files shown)


GarySmith:
Note: if you want to try this out this code change, see the notes in the [JIRA defect](https://jira.suse.de/browse/SCRD-5747) for getting things setup

jeremy-moffitt:
> It seems to fix the problem, but of course after logging in, we are not attempting to go back to the page that the user was on. Perhaps that can be addressed in another change, if that functionality is still desired.

Ideally that would be the behavior. I played around with the history goBack() and go() functions a bit without much success, but I think at this point its more important not to have the double login

jeremy-moffitt:
@guglo were you running minimized or in dev mode? if you run via run_dist.sh you should see the different behavior... in dev mode the error is replaced by a CORS error that I didnt work on since customers wont see it

guglo:
@jeremy-moffitt thanks, will try again.

PR #315, opened 2019-02-20
Added encrypt key handling for package page (SCRD-7652)

ghost:
1.When load the package page, it will prompt user for encrypt key if
the cloud config is encrypted.
2.The encrypt key is saved in cache, refresh the page, it will prompt user
for encrypt key again if cloud config is encrypted.


guglo:
It goes with https://gerrit.suse.provo.cloud//#/c/5673

src/localization/bundles/en.json
```diff
@@ -71,7 +71,7 @@
     "id.with_colon": "ID:",
     "common.encryptkey.message": "Please provide encryption key for running playbooks",
     "common.encryptkey.title": "Provide encryption key",
-    "common.encryptkey.description": "The operation you are about to perform requires the encryption key. Please provide the encryption key.",
+    "common.encryptkey.description": "The operation you are about to perform requires the encryption key. Please provide the encryption key. If provide wrong encryption key, operation will fail. Refresh the page to invoke this dialog again in case wrong key was provided. ",
```
GarySmith: The wording here needs a little fine-tuning. Consider:

The operation you are about to perform runs a playbook that requires the encryption key. If you provide wrong encryption key, the playbook will fail and you will need to refresh your browser window in order to re-enter the encryption key.

GarySmith (changes requested):

src/pages/ArdanaPackages.js
```diff
@@ -29,13 +31,42 @@ class ArdanaPackages extends Component {
...
     };
   }
 
-  componentWillMount() {
+  async componentWillMount() {
+    let isEncrypted = await isCloudConfigEncrypted();
+    if(isEncrypted) {
+      // use the window to save it in cache so it will
+      // not prompt every time when access this page
+      if(isEmpty(window.encryptKey)) {
```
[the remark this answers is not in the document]
guglo: @GarySmith  ok will give it a try.

src/pages/ArdanaPackages.js
```diff
@@ -29,13 +31,42 @@ class ArdanaPackages extends Component {
...
+        this.setState({showEncryptKeyModal: true});
+      }
+      else {
```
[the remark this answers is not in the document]
guglo: ok will update

src/localization/bundles/en.json
guglo: ok, will update. Thanks for the wordings fixes.

guglo:
To test is_encrypted API call in development env, from ardana-service dir
Not sure where to put this instruction permanently though.

src/localization/bundles/en.json
```diff
@@ -71,7 +71,7 @@
     "id.with_colon": "ID:",
     "common.encryptkey.message": "Please provide encryption key for running playbooks",
     "common.encryptkey.title": "Provide encryption key",
-    "common.encryptkey.description": "The operation you are about to perform requires the encryption key. Please provide the encryption key.",
+    "common.encryptkey.description": "The operation you are about to perform runs a playbook that requires the encryption key. If you provide the wrong encryption key, the playbook will fail and you will need to refresh your browser window in order to re-enter the encryption key.",
```
dotNomad: I think this can be written a bit more generally that doesn't put as much blame on the user. Perhaps:

> The operation you are about to perform runs a playbook that requires the encryption key. If the encryption key provided is incorrect the playbook will fail and the browser window will need to be refreshed in order to re-enter the encryption key.

dotNomad (approved):
A quick rewording suggestion. Looks good though, and works as intended. :+1:

src/localization/bundles/en.json
guglo: @dotNomad  will rewording it. thanks.

GarySmith (approved):

GarySmith:
LGTM, but need to be able to re-use this encryptKey throughout the application. Will create a defect for that.

PR #316, opened 2019-02-22
Added retry and cancel handling for replace compute host (SCRD-6817 SCRD-7848)

ghost:
The commit includes:

1. When failed during replace compute, retry and cancel buttons
show up. When user clicks retry, it will show a confirm dialog and
retry the failed playbook or action. When user clicks cancel, it
will a confirm dialog and goes back the server summary.
2. Added retry for activate, deactivate and delete. However
those action will not include cancel when playbook fails since
it only has a single page.


guglo:
This one depends on https://github.com/ArdanaCLM/ardana-installer-ui/pull/304. The stupid gihub include all the parent changes. My git seems to having issues lately that I got a couple of object corrupted errors so push it to secure the changes just in case.

guglo:
This PR is more of for store the code in repo so my local corrupted git won't cause problem. Will start a clean PR since the parent has been merged.

PR #317, opened 2019-02-26
Retry/cancel when replace/activate/deactivate/delete a compute (SCRD-6817 SCRD-7848)

ghost:
The commit includes:

1. When failed during replace compute, retry and cancel buttons
   show up. When user clicks retry, it will show a confirm dialog and
   retry the failed playbook or action. When user clicks cancel, it
   will show a confirm dialog and go back to the server summary.
2. Added retry for activate, deactivate and delete. However
   they will not have cancel when playbook fails since
   it only has a single page.
3. Fixed an issue in deactivate compute when there are instances to be migrated.
4. Prompt user for password when delete compute (SCRD-4442).
5. Make migration monitoring timeout to be 60 minustes.
6. Minor cleanups.
7. Fixed an issue when old server has instances and reachable


guglo:
Need one more around of real testing after rebase and dev testing.

src/UpdateWizard.js
```diff
@@ -265,7 +264,7 @@ class UpdateWizard extends InstallWizard {
           play.status = '';
         }
       });
-      this.updateGlobalState('playbookStatus', playStatus);
+      await  this.updateGlobalState('playbookStatus', playStatus);
```
jeremy-moffitt: nitpick, double spaces here... might need to lint this

src/localization/bundles/en.json
```diff
@@ -714,9 +716,15 @@
...
+    "server.replace.compute.failure.delete.cancel.confirm": "Cancelling could potentially leave the process of deleting the old compute server {0} in a half state. It will close the window, and any error messages will no longer be available to view.\n\nDo you want to cancel?",
+    "server.replace.compute.failure.disable.cancel.confirm": "Cancelling could potentially leave the process of disabling the old compute server {0} in a half state. It will close the window, and any error messages will no longer be available to view.\n\nDo you want to cancel?",
```
jeremy-moffitt: I'm not sure that the meaning of "in a half state" is clear here... would be curious about alternatives "partially complete" or "in an incomplete state" ... I dont know...

src/pages/UpdateServers.js
```diff
@@ -462,21 +497,53 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+        loading: true,
+        id: id,
+        osPassword: ''
       }
     });
+    try {
+      let promises = [getReachability(server['ip-addr'])];
+      let [conectivityStatus] = await Promise.all(promises);
+      this.setState(prev => ({
+        serverStatuses: {
```
jeremy-moffitt: its been a while since I mucked around with setting state, is it really necessary to explicitly set the statuses, ID, and the internal states just to set the isReachable state in the internal object? seems like there might be a more straightforward way to do this...

jeremy-moffitt (commented):
generally speaking this seems to work... I've got some concerns about some of the wording, and I wonder if theres a cleaner way to update the state in one spot

src/UpdateWizard.js
guglo: @jeremy-moffitt  ok will fix

src/localization/bundles/en.json
guglo: What I try to say here is when it failed in the middle of running deployment,  for example site.xml --limite...depending where it failed,  instead of fixing the problems and retry, click cancel can potentially leave the cloud in a wired situation.

src/pages/UpdateServers.js
guglo: I just did a quick fix for the problem in delete compute.  I picked up the code from deactivate so that at least they are in consistent code style. I am sure it could be made to be straight forward...however this seems to work at this point.

src/localization/bundles/en.json
```diff
@@ -714,9 +716,15 @@
...
+    "server.addserver.installos.cancel.confirm": "This will cancel the OS installation for the new compute server(s).  \n\nDo you want to cancel?",
+    "server.replace.compute.osinstall.failure.cancel.confirm": "Cancelling wil close the window and any error messages will no longer be available to view. \nAfter cancel, the new and undeployed compute server {0} shows up in servers Summary page, go to Add server page and unassign it from the role before replace compute server {1} again.  \n\nDo you want to cancel?",
```
jeremy-moffitt: the middle part of this should read:
`After canceling, the new and undeployed compute server {0} will show up in the Server Summary page. Navigate to the Add Server page to unassign it from it's role before attempting to replace server {1} again`

jeremy-moffitt: It might also be good to replace `Do you want to cancel?` with `Proceed with cancellation?`

jeremy-moffitt: also `will` is mis-spelled `wil` here

guglo: @jeremy-moffitt  ok , will update

jeremy-moffitt (approved):

eapodaca (changes requested):

src/pages/UpdateServers.js
```diff
@@ -462,21 +497,53 @@ class UpdateServers extends BaseUpdateWizardPage {
...
-        loading: false,
-        id: id
+        loading: true,
+        id: id,
+        osPassword: ''
       }
     });
+    try {
+      let promises = [getReachability(server['ip-addr'])];
+      let [conectivityStatus] = await Promise.all(promises);
```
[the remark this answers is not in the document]
guglo: @eapodaca cool will fix

src/pages/ReplaceServer/DeleteCompute.js
```diff
@@ -58,18 +58,28 @@ class DeleteCompute extends BaseUpdateWizardPage {
         fetchJson('/api/v2/server?source=sm,ov,manual'),
         fetchJson('/api/v2/cobbler')
       ];
+      let hasCobblerServer = false;
       const [ servers, cobblerStatus ] = await Promise.all(promises);
+      // if cobbler is present go find the old compute
+      if(cobblerStatus.cobbler) {
+        const [cobblerServers] = await Promise.all([fetchJson('/api/v2/cobbler/servers')]);
```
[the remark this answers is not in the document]
guglo: ditto

eapodaca (approved):

PR #319, opened 2019-02-26
Update styling on migration monitor progress

eapodaca:
* Use primary green color for progress bar
* Show spinner until 100%
* use black text color < 5% progress


guglo:
@eapodaca

PR #320, opened 2019-02-27
Refer to docs when controller replace fails (SCRD-3497)

GarySmith:


src/localization/bundles/en.json
```diff
@@ -600,7 +600,7 @@
     "server.replace": "Server Replacement in Progress",
     "server.replace.complete.heading": "Completed Server Replacement",
     "server.replace.complete": "Completed replacing server {0}",
-    "server.replace.prepare.failure": "Failed to prepare for replacing server. {0}",
+    "server.replace.controller.failure": "Failed to replace controller.  Check the \"Replacing a Controller Node\" section of the SUSE OpenStack Cloud Operations guide for troubleshooting.",
```
guglo: @GarySmith since the code passes the  error in, need  {0} here to show the error otherwise the message shows server.replace.controller.failure instead.

PR #322, opened 2019-03-01
Update bootstrap to 4.3.1 due to XSS security issue (SCRD-8112)

jeremy-moffitt:
Update bootstrap due to XSS vulnerability. 

A sanity check found no obvious visual differences between 4.1.2 and 4.3.1 but should be double-checked


jeremy-moffitt:
Please review as time permits, but wait to merge (approvals are fine, just don't merge it until I can update the dependency tarball in parallel)

PR #323, opened 2019-03-05
Consolidate to get and set the encryptKey (SCRD-8061)

ghost:
This commit includes:

1. Added a global var gEncryptKey and a set of
helper functions getCachedEncryptKey and setCachedEncryptKey.

2. The gEncryptKey will be shared among pages where it needs
to run playbooks with encryptKey.

3. Updated replace controller, replace compute, add compute,
activate/deactivate/delete compute to use helper functions
when it needs to deal with encryptKey.

4. Updated service info, package and configuration to use
helper functions when it needs to deal with encryptKey

### DIFF
--- a/src/UpdateWizard.js
+++ b/src/UpdateWizard.js
@@ -67,8 +67,6 @@ class UpdateWizard extends InstallWizard {
       // Indicate if the cloud configuration isEncrypted. If it is true,
       // it will force user to input encryptKey.
       isEncrypted: false,
-      // encryptKey from user input if isEncrypted is true otherwise it is empty.
-      encryptKey: ''
     };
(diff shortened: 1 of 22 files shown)


jeremy-moffitt:
tried without the encryption key and this seems to be fine, need to test with it enabled (or at least faked)

guglo:
To test is_encrypted API call in development env, from ardana-service dir
Not sure where to put this instruction permanently though.

jeremy-moffitt (approved):
tried this out on my local dev system with the encryption key setting enabled, behaved as expected

eapodaca (approved):
LGTM

PR #324, opened 2019-03-06
Hide undeployed servers in the servers page in day2 production (SCRD-6682)

ghost:
…6682)

### DIFF
--- a/src/components/CollapsibleTable.js
+++ b/src/components/CollapsibleTable.js
@@ -25,6 +25,7 @@ import {
 } from '../utils/ModelUtils.js';
 import { loadServerDiskUtilization } from '../utils/MonascaUtils.js';
 import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+import { isProduction } from '../utils/ConfigHelper.js';
 
 class CollapsibleTable extends Component {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


eapodaca (approved):
LGTM

jeremy-moffitt (changes requested):
The code looks fine, but when I tried this out, I still see the "replace" action on the compute server I added but haven't deployed yet

guglo:
@jeremy-moffitt you should not see any changes in dev, hide is only for production.  Just need to leave some convenient for dev testing.

guglo:
If force production in dev, when just drag a server in compute without run deploy...that server will NOT show up in server summary page day2.  however in day0, it should show up. 
 export function isProduction() {
  if (PRODUCTION) { // eslint-disable-line no-undef
    return true;
  }
  return true;
}

If NOT force production, then it will show up in the server summary in day2, no restriction there.

jeremy-moffitt:
@guglo sorry for the delay, trying it out now in production mode

jeremy-moffitt (approved):

jeremy-moffitt:
I see the filter applying in production mode (the server does appear, but is quickly filtered out), thanks!

PR #326, opened 2019-03-08
Update layout of yml editor (SCRD-7642)

eapodaca:
Update the layout of the yml editor to show same alert as config list.

### DIFF
--- a/src/pages/ValidateConfigFiles.js
+++ b/src/pages/ValidateConfigFiles.js
@@ -38,6 +38,19 @@ const TAB = {
   CONFIG_FORM: 'CONFIG_FORM'
 };
 
+
+const ErrorAlert = (props) => (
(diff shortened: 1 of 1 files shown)


src/pages/ValidateConfigFiles.js
```diff
@@ -82,38 +97,31 @@ class EditFile extends Component {
...
+                  inputValue={this.state.contents}
+                  inputName='fileContents'
+                  inputType='textarea'
+                  inputValidate={YamlValidator}
+                  inputAction={this.handleChange}
+                />
+              </Otherwise>
+            </Choose>
+          </div>
+          <div className="col-md-6">
```
dotNomad: Feels like this div should also be conditionally rendered? If there isn't an error we just get an empty div.

eapodaca: Good point, updated.

PR #327, opened 2019-03-11
Show failed steps upon playbook cancel (SCRD-8020)

dotNomad:


src/components/PlaybookProgress.js
```diff
@@ -262,6 +267,12 @@ class PlaybookProgress extends Component {
         }
       }
 
+      // if the global playbook status has failed then any in progress
+      // playbooks must also be failed
+      if (hasFailed && status === STATUS.IN_PROGRESS) {
+        status = STATUS.FAILED;
+      }
```
guglo: @dotNomad  This is more of for cancel case, if hasFailed, instead of making the step failed , just not show the loading button?

guglo: If user cancels (click the left side cancel, not the cancel button showing up when it failed) the playbook,  the step still shows in progress,  current behavior is after cancel intentionally the step has loading icon at the right side. With this change, the step doesn't have the loading icon but with red text, however cancelling doesn't mean the playbook has errors, just simply not continue. It might be better that eave the status as in progress as it is, just NOT showing the loading. Hope it makes sense. If want to have red text ...it is fine with me , just want to raise my concerns. 
I never tested cancel intentionally in the middle of the playbook running in the real environment, not sure what exactly the output of the playbook though.

jeremy-moffitt: I'm a bit torn on this ... the playbook is stopped, so we definitely want to ditch the loading indicator... I think continuing to use the same color we do when a playbook is running might hint that its still running, whereas red clearly means it is not... if we think of it as red = stopped (for whatever reason), gray = not started, black = running, green = finished, then I think this works

dotNomad: Not sure which is the best option after talking a bit with @guglo
- Show as `STATUS.FAILED` because we cancelled meaning the Playbooks most likely didn't complete. It sounds like they may have and they just haven't been reported yet. So may not be correct on the corner case.
- Show as `STATUS.IN_PROGRESS` but have a conditional to remove the loading indicator. May be the most accurate as a new status hasn't been reported, and removes the confusing loading indicator.
- Show as `STATUS.NOT_STARTED` reverting it to the light grey color - showing neither in progress or complete.

guglo: @dotNomad @jeremy-moffitt it is fine with red text for all the in progress steps when cancel intentionally as long as we have thought through.

PR #328, opened 2019-03-12
Select specific nodes to wipe disks in day0 (SCRD6836)

ghost:
This commit includes:
1. Added a transfer table to allow user deselect or select nodes when
wipe disks checked.
2. run site playbook and wipe disk playbook separately when
wipe disks checked.


dotNomad:
The multiple select is looking good. Would suggest adding spaces between each selected node though.
`controller1,controller2,controller3,ironic-compute-1` Looks like this right now.

guglo:
@dotNomad  decided to use transfer table instead.

jeremy-moffitt:
I'm a little confused on the workflow where I should see this... when I add 2 compute nodes and deploy it just kicks off... if I choose installOs it does show me the transfer box, is that what I should be concerned with?

guglo:
@jeremy-moffitt  it is in day0 , not in day2. You should see it review configure files page in day0 , where you can edit model, edit service template and set some deployment configure. In the deployment tab, if you check wipe disks , the transfer table will show up with node lists.

guglo:
image

jeremy-moffitt:
@guglo thanks for clarifying! I see it now. My one piece of feedback is I wonder if there would be better headings in this context than the "available servers" and "selected servers" ... Might have to ask around for some suggestions about what to use instead though... "Wipe Disk Servers" sounds odd and "Servers whose disks will not be wiped" is both too long and awkward... but I'm just not sure the context is clear enough to understand what "selected" and "available" mean.

guglo:
@jeremy-moffitt  ok thanks for the advice. I will ask around and make some wording changes.

guglo:
image

dotNomad (approved):
Transfer table works as intended it looks like. The warning and improved tooltip are also helpful. :+1:

guglo:
Tested deploying on 3 nodes. Found a big issue that I don't have cp_internal/CloudModel.yaml to checked for the ansible host names until configure processor and ready to deployment step finished. I Will add an action step after that step so I can exact names ....

guglo:
Tested on 3 nodes system. Fixed a couple of issues.

eapodaca (approved):
LGTM.

PR #332, opened 2019-03-19
Load the model before service statuses (SCRD-8483)

eapodaca:
- Wait for model to be loaded before fetching server statuses.
- If there is a pending promise to load the model, return that promise.
Can potentially dedupe simultaneous requests.

### DIFF
--- a/src/InstallWizard.js
+++ b/src/InstallWizard.js
@@ -329,13 +329,26 @@ class InstallWizard extends Component {
   // Pages within the installer may request that the model be forceably loaded
   // from disk, espcially when a change is made to directly to the model files
   // to the model.  Returns a promise
-  loadModel = () => fetchJson('/api/v2/model')
-    .then(responseData => {
-      this.setState({'model': fromJS(responseData)});
-    })
-    .catch((error) => {
-      console.log('Unable to retrieve saved model');// eslint-disable-line no-console
-    })
+  loadModel = () => {
+    // If there is a pending promise to fetch the model return that promise.
+    if(!this.fetchModelPromise) {
+      this.fetchModelPromise = new Promise((resolve, reject) => {
+        fetchJson('/api/v2/model')
+          .then(responseData => {
+            this.setState({'model': fromJS(responseData)}, () => {
+              this.fetchModelPromise = undefined;
+              resolve();
+            });
+          })
+          .catch((error) => {
+            console.log('Unable to retrieve saved model');// eslint-disable-line no-console
+            this.fetchModelPromise = undefined;
+            reject(error);
+          });
+      });
+    }
+    return this.fetchModelPromise;
+  }
 
   // Pages within the installer may request that the model be saved to disk,
   // which is especially important when some significant change has been made
--- a/src/pages/UpdateServers.js
+++ b/src/pages/UpdateServers.js
@@ -124,15 +124,16 @@ class UpdateServers extends BaseUpdateWizardPage {
           };
         });
       }
+      let internalModel = this.state.internalModel;
       if(!this.state.internalModel) {
-        const model = await getInternalModel();
-        this.setState({ internalModel: fromJS(model) });
+        internalModel = fromJS(await getInternalModel());
+        this.setState({ internalModel });
       }
       if(!this.state.osUsername) {
         this.getUsername();
       }
       let promises = [
-        this.getServerStatuses(),
+        this.getServerStatuses(internalModel),
         this.checkMonasca()
       ];
       await Promise.all(promises);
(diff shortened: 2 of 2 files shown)


src/pages/UpdateServers.js
```diff
@@ -254,6 +254,10 @@ class UpdateServers extends BaseUpdateWizardPage {
   }
 
   async getServerStatuses() {
+    //we need the model from the props here, we will try to wait for them to show up
+    while(!this.props.model || !this.state.internalModel) {
```
GarySmith: We should not be checking for `this.props.model` here.  The `setState` call that we are waiting for does not populate this.props -- react populates this before `componentDidMount` is called.

src/pages/UpdateServers.js
```diff
@@ -254,6 +254,10 @@ class UpdateServers extends BaseUpdateWizardPage {
...
+      await new Promise((resolve, reject) => setTimeout(resolve, 10));
```
GarySmith: Polling for completion is an ugly way to handle this, and is going to be problematic if there is a possibility for some exception occurring that can prevent this from happening.

Since this function is called in only one place, componentDidMount, how about just passing in the model as a function parameter directly; we really don't need to even look in the state here, do we?

It looks like the `checkMonasca` also has all sorts of contortions to try to work around this same problem that could be handled in a much more straightforward way.

src/pages/UpdateServers.js
eapodaca: This issue is happening because the code attempts to perform some computation on `this.props.model` which is not populated in this component. I think what we need to do is pass in a promise for when the model will be loaded so that I can await for it.

src/pages/UpdateServers.js
```diff
@@ -254,6 +254,8 @@ class UpdateServers extends BaseUpdateWizardPage {
   }
 
   async getServerStatuses() {
+    // make sure the model it loaded, we need to fetch some things from it.
+    await this.props.loadModel();
```
GarySmith: This still does not solve the race condition -- the call to `this.setState` on line 337 is not guaranteed to complete before the promise is resolved, so line 259 below can still be attempting to reference `this.state.internalModel` before it is set. The conditional `?.` will avoid dereferencing it in that case, so now instead of having a race condition that causes and error we have a race condition that causes no status to be returned.

src/InstallWizard.js
```diff
@@ -329,13 +329,21 @@ class InstallWizard extends Component {
...
-  loadModel = () => fetchJson('/api/v2/model')
-    .then(responseData => {
-      this.setState({'model': fromJS(responseData)});
-    })
-    .catch((error) => {
-      console.log('Unable to retrieve saved model');// eslint-disable-line no-console
-    })
+  loadModel() {
+    // If there is a pending promise to fetch the model return that promise.
+    if(this.fetchModelPromise?.done) {
```
GarySmith: Shouldn't this be `! this.fetchModelPromise?.done` ? This check would seem to cause the code to be executed even if it has already been resolved.

eapodaca: No. When this method is called the caller wants the model to be loaded, so we will load it. If there is a existing promise that has not finished we will not create another request to load the model. We are already loading the model.

eapodaca: That being said there may have been a logic error in there, so I have made that a little more explicit.

src/pages/UpdateServers.js
eapodaca: The promise that loadModel now returns does not resolve until react gives us a callback from setState. I didn't quite understand what you were referring to the first go around, this is I think a better solution.

src/InstallWizard.js
```diff
@@ -329,13 +329,26 @@ class InstallWizard extends Component {
...
-  loadModel = () => fetchJson('/api/v2/model')
-    .then(responseData => {
-      this.setState({'model': fromJS(responseData)});
-    })
-    .catch((error) => {
-      console.log('Unable to retrieve saved model');// eslint-disable-line no-console
-    })
+  loadModel = () => {
+    // If there is a pending promise to fetch the model return that promise.
+    if(!this.fetchModelPromise || this.fetchModelPromise?.done) {
```
GarySmith: This still looks like it will run another fetch if the first one is still pending. I think it would be better to just use:
```
if (!this.fetchModelPromise) {
   this.fetchModelPromise = ... 
       (same stuff you have)
}
return this.fetchModelPromise;
```
***except*** remove assigning and checking for the `done` field

eapodaca: With that logic this method will only be able to fetch the model once. Any subsequent calls will get the promise that was created the first time. What we could do, unset `fetchModelPromise` when it is resolved, instead of relying on the done flag.

eapodaca:
@GarySmith I've updated the `getServerStatuses` method to take the model as a parameter.

src/pages/UpdateServers.js
```diff
@@ -124,15 +124,16 @@ class UpdateServers extends BaseUpdateWizardPage {
           };
         });
       }
+      let model;
```
GarySmith: There is still a problem here: if this.state.internalModel is already populated, then model will remain undefined, and line 136 will be passing undefined. I think the line above probably needs to be `let model = this.state.internalModel;`

eapodaca: Good catch, we can use that as the default value for the model to address this.

GarySmith:
Note that I did not test this, but I'm Jeremy will before he approves it :-)

jeremy-moffitt:
I'm seeing the same error that Jordan is

eapodaca:
That is what I get for pretending I know what I am doing. Fixed now.

PR #333, opened 2019-03-20
Updated steps necessary for deactivating/deleting a compute node (SCRD-8470)

ghost:
This commit includes:

1. Add steps when delete a  compute.
2. Make sure actions are not available for non compute nodes when
operation is in progress in the Server Summary.
3. Move up the step of disabling compute service in deactivate a compute
4. Handle instances evacuation if the compute node is not reachable when deactivate a compute. 
5. Moved some action names to the constants.js


eapodaca (approved):
This looks mostly OK. I have not tested.

src/localization/bundles/en.json
```diff
@@ -686,14 +686,16 @@
...
+    "server.deactivate.confirm.message_instances_evacuate": "Are you sure you want to deactivate this compute server, {0} it is not reachable and currently hosting {1} instance(s) which will need to be evacuated? Make sure the compute server is configured with block storage or shared storage.",
     "server.activate.text": "Activate {0}",
     "server.activate_service.text": "Add {0} to Nova scheduler",
     "server.deactivate.text": "Deactivate {0}",
     "server.retreive.serverstatus.error": "Failed to retrieve compute server status: {0}",
     "server.migrate_title": "Migrate instances from {0}",
     "server.migrate.prompt": "Select a computer host to migrate the instances on {0} host to.",
-    "server.deactivate.confirm.message_instances_cant_be_migrated": "Are you sure you want to deactivate this compute server ({0})? It is currently hosting {1} instance(s) which may be terminated for made in accessible by this action.",
+    "server.evacuate.prompt": "Select a computer host to evacuate the instances on {0} host to.",
```
GarySmith: I believe this is supposed to say "compute host" instead of "computer host"

src/localization/bundles/en.json
```diff
@@ -686,14 +686,16 @@
...
+    "server.deactivate.confirm.message_instances_terminate": "Are you sure you want to deactivate this compute server ({0})? It is currently hosting {1} instance(s) which may be terminated for made in accessible by this action.",
```
GarySmith: "for made" is supposed to be "or made"

src/pages/ManageComputeHost/DeactivateComputeHost.js
```diff
@@ -55,17 +48,17 @@ class DeactivateComputeHost extends DisableComputeServiceNetwork {
...
-      playbooks.push({
-        name: MIGRATE_INSTANCES,
-        action: ::this.migrateInstances
-      });
+    let playbooks = [{
+      name: constants.DISABLE_COMPUTE_SERVICE_ACTION,
+      action: ::this.disableCompServices
+    }];
+
+    let instancesPlaybooks = this.getPlaybooksForInstances();
```
GarySmith: Ditto above, these 4 lines should just be `playbooks = playbooks.concat(this.getPlaybooksForInstances());`

src/pages/ReplaceServer/DisableComputeServiceNetwork.js
```diff
@@ -363,34 +351,39 @@ class DisableComputeServiceNetwork extends BaseUpdateWizardPage {
...
+  getSteps() {
+    let steps = [{
+      label: translate('server.deploy.progress.disable_compute_service'),
+      playbooks: [constants.DISABLE_COMPUTE_SERVICE_ACTION]
+    }, {
+      label: translate('server.deploy.progress.remove_from_aggregates'),
+      playbooks: [constants.REMOVE_FROM_AGGREGATES_ACTION]
+    }];
+
+    let instancesSteps = this.getStepsForInstances();
```
GarySmith: Ditto

src/pages/ManageComputeHost/DeleteComputeHost.js
```diff
@@ -14,9 +14,143 @@
...
+      '/api/v2/compute/aggregates/' + this.props.operationProps.oldServer.hostname;
+    logger('DELETE ' + apiUrl);
+    return deleteJson(apiUrl)
+      .then((response) => {
+        const logMsg =
+          'Got response from removing aggregates for compute host ' +
+          this.props.operationProps.oldServer.hostname;
+        logProgressResponse(logger, response, logMsg);
+      })
+      .catch((error) => {
```
GarySmith: This error handling logic is pretty convoluted, and seems to be copy/pasted below. It's unclear why a failure dialog is only presented for some 500 errors and not others. Or why some messages are translated and others are not, or why some errors cause an error to be thrown and others do not.

src/pages/ReplaceServer/DisableComputeServiceNetwork.js
```diff
@@ -406,10 +399,31 @@ class DisableComputeServiceNetwork extends BaseUpdateWizardPage {
...
+        return this.disableCompServices(logger);
+      })
+    }, {
+      name: constants.REMOVE_FROM_AGGREGATES_ACTION,
+      action: ((logger) => {
+        return this.removeAggregates(logger);
+      })
+    }];
+
+    let instancesPlaybooks = this.getPlaybooksForInstances();
```
GarySmith: Ditto

GarySmith (commented):
Lots of small stuff

src/pages/ManageComputeHost/DeleteComputeHost.js
guglo: @GarySmith  it is copied and pasted. I tried to move them to a common place like util, but gave it up , the effort seems to be more involved than I initially thought. 
When delete aggregates, since the backend deletes host aggregates one by one, if there are errors, we don't stop, in the end if the results include good and bad responses, backends returns response with 500 status code. Then UI checks if error code is 500 and have failed (bad) and have deleted(good), pops a confirm dialog to ask user to continue or not.

guglo: Talked to Gary, I will add some comments so we will be a little bit clear about the error handling part.

dotNomad (approved):
Tested and seems to be working as intended. Both steps (removing aggregates and disabling network agents) appear when deleting a disabled compute node.

jeremy-moffitt:
> Tested and seems to be working as intended. Both steps (removing aggregates and disabling network agents) appear when deleting a disabled compute node.

How did you simulate the delete? When I try deactivating in a dev environment the node just goes back to active since there's not a real backend behind it

guglo:
@jeremy-moffitt in the ardana-service tools/compute-mock-data.json "compute_services_status": {
    "nova-compute": false
  },
refresh delete will show up
if you want to test evacuate step, add some in get_instances
"get_instances": [{
    "id": "8279e65d-6e87-4a50-b789-96edd753fbb2",
    "name": "test3"
  }, {
    "id": "1d51f18f-27fd-4c34-a0aa-c07a5e9462e7",
    "name": "test2"
  }]

jeremy-moffitt (approved):
tried it out with the suggested tweaks to make the compute servers deletable, seemed fine... not merging though pending confirmation from @GarySmith that his concerns have been addressed

jeremy-moffitt (approved):
re-tested with the latest change, still seems to work ok. One thing I found a little odd, the first time you open the delete dialog, there is a comparatively lengthy "loading please wait..." that I think was checking about a password or something (the password prompt is what shows once it loads). The timing is a bit odd, but it works

src/pages/UpdateServers.js
```diff
@@ -688,7 +688,7 @@ class UpdateServers extends BaseUpdateWizardPage {
             {translate('server.deactivate.confirm.message_instances_evacuate', id, instances.length)}
           </If>
           <If condition={!loading && server.hasInstances && otherHosts.length === 0}>
-            {translate('server.deactivate.confirm.message_instances_terminate', id, instances.length)}
+            {translate('server.deactivate.confirm.message_instances_terminate', id, instances.length, id)}
```
jeremy-moffitt: couldnt you just reuse {0} or will the string replacement not work that way?

jeremy-moffitt (approved):
thanks for fixing the repeat parameter issue!

PR #334, opened 2019-03-20
Show status playbook only if service has one (SCRD-5829)

dotNomad:


dotNomad:
A playbook can be added in a development environment by navigating to `../data/ardana-ansible` in `ardana-service` and creating a file with the playbook's name you'd like to add.
For example I created `cinder-status.yml` and `heat-status.yml` to test this change.

jeremy-moffitt:
so there is a fake keystone-status log that streams if you run the keystone status playbook without this change.... with this change that playbook doesnt seem to run... possibly because my the keystone-status.yml playbook I created was goobbledygook... not sure...

dotNomad:
Good catch @jeremy-moffitt I'll investigate

dotNomad:
Case sensitivity issue... :man_facepalming:
Should now be able to see the keystone-status log since the status playbooks are now correctly named, with no capitalization.

PR #335, opened 2019-03-21
Add button to copy server details (SCRD-8243)

eapodaca:
* Add button to copy server details from server to be replaced.
* Refactor some common copy code from available servers functionality.

### DIFF
--- a/src/components/ReplaceServerDetails.js
+++ b/src/components/ReplaceServerDetails.js
@@ -224,28 +224,30 @@ class ReplaceServerDetails extends Component {
     });
   }
 
+  copyFromServer(server, inputValue, isValid) {
+    // Copy the fields of interest from known servers, and set the validity
+    // of each field depending on whether it is populated
+    for (let key of this.getInputNames()) {
+      const valueToCopy = server[key] || '';
+      inputValue = inputValue.set(key, valueToCopy);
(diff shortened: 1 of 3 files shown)


jeremy-moffitt:
Seems to work ok... still deciding if the button in the top right is the best "affordance" to use for this... but not sure a better alternative at the moment

eapodaca:
I wasn't sure about that either, looks equally strange on the left also.

guglo:
@eapodaca  replacing a compute node is to add a new compute node and delete the old compute . If user enter the information manually ...all the information will be saved into manually added servers. If replace failed, the server will be in manually added server and if user want to try again, they can use select from available servers to pick up the info they entered.  If still want to have this copy button in replace compute dialog, it needs to coordinate with select from available servers and make sure some fields have no conflict with the existing ones and no copy of password. 

In replace a controller, it only replaces IPMI stuffs, not sure the info will be saved into manual servers. If want to want to copy from existing ones, it needs to coordinate with select from available servers and make sure some fields have no conflict with the existing ones and no copy of password. 

At this point I think it is not a good idea to have the copy button there.

eapodaca:
Alright if there is a way for users to move forward and not get stuck I am fine with this. Closing.

PR #336, opened 2019-03-25
Prevent manipulating servers during op progress (SCRD-8553)

ghost:
Prevent editing/deleting/moving servers with roles in AddServer when replacing a server in progress

### DIFF
--- a/src/components/ServerTable.js
+++ b/src/components/ServerTable.js
@@ -29,11 +29,14 @@ class ServerTable extends Component {
         // when it is in addserver mode, if we have a list of deployed servers
         if(this.props.isUpdateMode &&
            this.props.deployedServers && this.props.deployedServers.length > 0) {
-          // if the server item is NOT in the deployedServers, will present
+          // If the server item is NOT in the deployedServers, will present
           // editAction and deleteAction
-          if(!this.props.deployedServers.some(server => {
+          // Or if isSafeMode is false, will present editAction and deleteAction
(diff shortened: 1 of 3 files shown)


eapodaca (approved):
LGTM. Did not test.

GarySmith (approved):
Ditto. LGTM but did not test

dotNomad (changes requested):
While replacing a compute server was able to navigate to Add Server, but I noticed a few issues.

- If I manually add a new server and drag it over, the previously deployed compute servers become read only. If I then deploy the other servers are still stuck in read only mode. I can't seem to find a way to unlock them so they can be edited or deleted again.
Upon further testing this appears to happen if just importing servers from a CSV file.

- While the dragging is disabled it is possible to still drag the servers by highlighting their text, and dragging the highlighted text. This is fairly easy to "accidentally" do if you expect them to drag as when you do and they are disabled the text will be highlighted.

jeremy-moffitt (approved):
tried it out, this worked fine after learning how to test it :)

dotNomad:
Sounds like it is expected to have previously deployed compute servers be read only. Approving, and opening a separate issue for the dragging highlighted text bug.

dotNomad (approved):

guglo:
Talked Jordan , the deployed servers are not allowed to edit and delete by default in Add Servers, unless user uncheck the safe mode. Regarding the drag and drop, it is more of the drag and drop component issue, Jordan will file a defect for that issue.

PR #339, opened 2019-03-29
Prevent drag of highlighted server text (SCRD-8585)

jeremy-moffitt:
Prevents server row highlights from being dragged as part of drag and drop.

### DIFF
--- a/src/styles/common.less
+++ b/src/styles/common.less
@@ -144,6 +144,13 @@ code {
   cursor: move;
 }
 
+.notDraggable {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  user-select: none;
+  -ms-user-select: none;
+}
+
 .button-box {
   vertical-align: middle;
   display: inline-block;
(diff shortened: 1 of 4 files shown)


src/styles/common.less
```diff
@@ -144,6 +144,12 @@ code {
   cursor: move;
 }
 
+.notDraggable {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  user-select: none;
```
dotNomad: Looks like we will need to add `-ms-user-select: none;` for IE and Edge.

src/styles/components/serverutils.less
```diff
@@ -46,6 +46,9 @@
     color: @detailcolor;
     position: relative;
     padding: 10px;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    user-select: none;
```
dotNomad: `-ms-user-select: none;` again

jeremy-moffitt:
made suggested changes to apply the highlighting restriction to MS browsers as well

PR #342, opened 2019-04-01
Add SES integration into day0/2 (SOC-8555)

eapodaca:
* Adds ability to upload ses_config.yml on configuration screens
* Adds required playbooks to reconfigure steps

### DIFF
--- a/src/pages/ServiceConfiguration.js
+++ b/src/pages/ServiceConfiguration.js
@@ -119,7 +119,7 @@ class ServiceConfiguration extends Component {
   }
 
   getPlaybooks = () => {
-    let playbooksToRun = {};
+    let playbooksToRun = {playbooks: []};
     const changedServices = this.serviceTemplatesTab.getChangedServices();
     if (changedServices?.length > 0) {
       playbooksToRun.steps = [{
@@ -129,10 +129,18 @@ class ServiceConfiguration extends Component {
         label: translate('deploy.progress.ready-deployment'),
         playbooks: [constants.READY_DEPLOYMENT_PLAYBOOK + '.yml']
       }];
+      playbooksToRun.playbooks.push(constants.PRE_DEPLOYMENT_PLAYBOOK);
+      if(this.state.sesEnabled) {
+        playbooksToRun.steps.push({
+          label: translate('deploy.ses'),
+          playbooks: [`${constants.DEPLOY_SES_PLAYBOOK}.yml`]
+        });
+        changedServices.push('cinder', 'glance', 'nova'); // we want to be sure ardana-reconfigure is triggered.
+        playbooksToRun.playbooks.push(constants.DEPLOY_SES_PLAYBOOK);
+      }
       const serviceName = changedServices[0];
       if (changedServices.length === 1 && UPDATEABLE_SERVICES.includes(serviceName)) {
-        playbooksToRun.playbooks = [
-          constants.PRE_DEPLOYMENT_PLAYBOOK, serviceName + '-reconfigure', serviceName + '-status'];
+        playbooksToRun.playbooks.push(serviceName + '-reconfigure', serviceName + '-status');
         playbooksToRun.steps.push({
           label: translate('deploy.progress.update'),
           playbooks: [serviceName + '-reconfigure.yml']
@@ -141,8 +149,7 @@ class ServiceConfiguration extends Component {
           playbooks: [serviceName + '-status.yml']
         });
       } else {
-        playbooksToRun.playbooks = [
-          constants.PRE_DEPLOYMENT_PLAYBOOK, constants.ARDANA_RECONFIGURE_PLAYBOOK];
+        playbooksToRun.playbooks.push(constants.ARDANA_RECONFIGURE_PLAYBOOK);
         playbooksToRun.steps.push({
           label: translate('deploy.progress.update'),
           playbooks: [constants.ARDANA_RECONFIGURE_PLAYBOOK + '.yml']
@@ -154,6 +161,10 @@ class ServiceConfiguration extends Component {
 
   toDisableUpdate = () => {
     return !this.state.isChanged || (this.state.isEncrypted && isEmpty(this.state.encryptKey));
+  }
+
+  setSesEnabled(sesEnabled) {
+    this.setState({ sesEnabled });
   }
 
   renderContent = () => {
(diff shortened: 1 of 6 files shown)


eapodaca:
Could really use some feedback on the changes I made to `src/pages/CloudDeployProgress.js`. The place in the order does not really feel right to me. Have any thoughts?

src/pages/ValidateConfigFiles/ServiceTemplatesTab.js
```diff
@@ -307,6 +307,31 @@ class ServiceTemplatesTab extends Component {
     );
   }
 
+  async handleSesUpload(file) {
+    this.setState({ loading: true });
+    let fileContents = await readFile(file);
+    await postJson('/api/v2/service/files/ses/ses_config.yml', JSON.stringify(fileContents));
```
GarySmith: I believe that `JSON.stringify` is not needed since `fileContents` is already a string.

src/pages/ValidateConfigFiles/ServiceTemplatesTab.js
```diff
@@ -307,6 +307,31 @@ class ServiceTemplatesTab extends Component {
...
+          if(serviceFile.changedFiles) {
+            serviceFile.changedFiles.push('ses_config.yml');
+          } else {
+            serviceFile.changedFiles = ['ses_config.yml'];
+          }
+        }
+        return serviceFile;
+      })
+    }));
+    this.props.hasChange?.(true);
```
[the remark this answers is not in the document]
eapodaca: If hasChange is a function call with those parameters.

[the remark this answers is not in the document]
eapodaca: compiles to something like:
```javascript
this.props.hasChange === 'function' && this.props.hasChange(true);
```

src/pages/ValidateConfigFiles/ServiceTemplatesTab.js
eapodaca: The way that other file content is posed in this file is the [same](https://github.com/ArdanaCLM/ardana-installer-ui/blob/23c9dc2e2e3876af914a897b36d7a06adb61f41f/src/pages/ValidateConfigFiles/ServiceTemplatesTab.js#L62). Removing the `JSON.stringify` call causes the `ardana-service` to return (400):
```
The browser (or proxy) sent a request that this server could not understand
```

src/pages/ValidateConfigFiles/ServiceTemplatesTab.js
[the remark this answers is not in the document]
GarySmith: ok, that makes sense. It looked like function invocation, but I didn't realize hasChange was a function (it doesn't sound like one). Next time I'll download the whole PR so that I can see the context better

src/pages/ValidateConfigFiles/ServiceTemplatesTab.js
GarySmith: Thanks, that's good to know.

GarySmith:
> Could really use some feedback on the changes I made to `src/pages/CloudDeployProgress.js`. The place in the order does not really feel right to me. Have any thoughts?

I'm not sure either, but I just poked around a little and found that ses-deploy.yml is already being called by the deploy playbooks for cinder, glance, nova, and swift, so it looks like you don't need to call it explicitly at all.

GarySmith:
replying in-line. Not sure if those will get posted without this comment.

eapodaca:
Alright, I've removed the explicit calls to `deploy-ses` in `CloudDeployProgress.js` it will get called properly in the regular deploy process.

eapodaca:
To test this on a local dev env extract to `data/my_cloud/config/` in the `ardana-service`.

[ses.zip](https://github.com/ArdanaCLM/ardana-installer-ui/files/3036180/ses.zip)

eapodaca:
here is an example `ses_config.yml` that can be uploaded:
```yml
ceph_conf:
 cluster_network: 10.84.56.0/21
 fsid: d5d7c7cb-5858-3218-a36f-d028df7b0673
 mon_host: 10.84.56.8, 10.84.56.9, 10.84.56.7
 mon_initial_members: ses-osd1, ses-osd2, ses-osd3
 public_network: 10.84.56.0/21
 cinder:
 key: AQCdfIRaxefEMxAAW4zp2My/5HjoST2Y8mJg8A==
 rbd_store_pool: mycloud-cinder
 rbd_store_user: cinder
 cinder-backup:
 key: AQBb8hdbrY2bNRAAqJC2ZzR5Q4yrionh7V5PkQ==
 rbd_store_pool: mycloud-backups
 rbd_store_user: cinder-backup
 glance:
 key: AQD9eYRachg1NxAAiT6Hw/xYDA1vwSWLItLpgA==
 rbd_store_pool: mycloud-glance
 rbd_store_user: glance
 nova:
 rbd_store_pool: mycloud-nova
 radosgw_urls:
     - http://10.84.56.7:80/swift/v1
     - http://10.84.56.8:80/swift/v1
```

src/pages/ServiceConfiguration.js
```diff
@@ -129,6 +129,17 @@ class ServiceConfiguration extends Component {
         label: translate('deploy.progress.ready-deployment'),
         playbooks: [constants.READY_DEPLOYMENT_PLAYBOOK + '.yml']
       }];
+      if(this.state.sesEnabled) {
```
guglo: @eapodaca  looks like the only way to set sesEnabled true is to upload file. Once my file is there after I  refresh my page ..then sesEnabled is not true anymore. When I make changes to ses_configure file, it doesn' get into this block.

guglo: After upload the file, I run the update, some of the steps are not updated.

guglo: image

eapodaca: This edge case causes problems even editing config files. If you edit a file and save it the update button is enabled. Refresh and its disabled, but the changes are not yet committed in the my_cloud repo.

guglo: ok, refreshing page after load the ses config file might not be a normal use case. I am ok with it. However  if I don't refresh page, it shows the deploy and reconfigure steps in the progress...but those steps are not updated.

guglo: @eapodaca  rerun the changes, now I understand why the steps are not updated. For Deploy SES step,  need to add the playbook into playbooksToRun.playbooks if his.state.sesEnabled. For reconfiguring cinder, glance, nova which I believe they are part of ardana reconfiguring, in the production I assume they are updated..however in development, they won't be updated unless you mock them.

guglo:
updated jira item to be SOC

PR #344, opened 2019-04-03
Make useful links persist (SCRD-8302)

dotNomad:
If the deployment page was refreshed the useful links in state
could be lost. Persisting them fixes this.

### DIFF
--- a/src/pages/CloudDeployProgress.js
+++ b/src/pages/CloudDeployProgress.js
@@ -23,6 +23,7 @@ import { PlaybookProgress } from '../components/PlaybookProgress.js';
 import { ErrorBanner } from '../components/Messages.js';
 import { getInternalModel } from './topology/TopologyUtils.js';
 
+const GET_EXTERNAL_LINKS = 'get_external_links';
 const GET_GEN_CLOUD_MODEL = 'get_generated_cloud_model';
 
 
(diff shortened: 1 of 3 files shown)


jeremy-moffitt:
tried this out on my dev environment, did a reload during the deployment process and got `Failed to check in-progress playbook status for get_external_links.yml - 0. NOT FOUND`  which doesnt seem tied to this specific change and is probably only because of the fake backend, but may be worth looking at for developer purposes

dotNomad:
@jeremy-moffitt Added a step called "Persisting endpoints" to the step list. Now there should be something in progress for the duration of the deployment.

src/pages/CloudDeployProgress.js
```diff
@@ -203,11 +209,18 @@ class CloudDeployProgress extends BaseWizardPage {
...
+      name: GET_EXTERNAL_LINKS,
       action: ((logger) => {
         return fetchJson('api/v2/external_urls')
           .then((response) => {
             this.props.updateGlobalState('usefulLinks', response);
+          })
+          .catch((error) => {
+            logger('Warning: Failed to fetch from api/v2/external_urls.');
+            logger('Unable to complete the ' +
+              `"${translate('deploy.progress.external-urls')}" step: ` +
```
jeremy-moffitt: this might be overkill (we dont usually translate log messages)

jeremy-moffitt: I should've refreshed the page

PR #345, opened 2019-04-05
Fixed issue when remove role and delete server (SCRD-8719)

ghost:
This commit includes a couple of fixes in day0 and day2:

1. Cleaned up the logic to update server correctly in manually
added servers and discovered servers list when remove role
of a server (drag server to the left) in day0 and day2
2. Used the common function to remove server from model to avoid the
problem where it deletes server with -1 index.
3. Fixed missing sessionKey when click save and continue button
without click Test first.
4. Fixed lost check possible null values of model in day0

### DIFF
--- a/src/localization/bundles/en.json
+++ b/src/localization/bundles/en.json
@@ -274,6 +274,7 @@
     "server.discover.get.error": "Failed to get discovered servers",
     "server.discover.update.error": "Failed to update discovered server {0}",
     "server.discover.delete.server.error": "Failed to delete server {0}",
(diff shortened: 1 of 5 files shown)


guglo:
Please check https://jira.suse.de/browse/SCRD-8719 for what to test.

jeremy-moffitt:
reproduced issue on master, tested fix and it works... still looking over the code

jeremy-moffitt (approved):
LGTM , I had to pick through the checkInputs change a bit, but looks good and solves the problem

eapodaca (approved):
LGTM. Works as expected.

PR #348, opened 2019-04-12
Added prepare deployment/deploy capability to model config (SCRD-8879)

ghost:
This commit incldues:

1. Add prepare deployment button to run config-processor-run and ready-deployemnt
   playbooks after validation.
2. Add deploy button to deploy model changes after validation.
3. New process pages should be able to deal with encrypt key, retry and refresh. When the
   process in progress, server summary page or add server page should not be able to run
   process.


src/localization/bundles/en.json
```diff
@@ -139,7 +139,8 @@
     "validate.config.files.heading": "Review configuration files",
     "validate.config.files.validate": "Validate",
     "validate.config.files.deploy": "Deploy",
-    "validate.config.files.msg.info0": "Validated model changes will be committed but are not automatically deployed. The model may be edited here in support of server add/replace use cases, or as needed to support service configuration changes. The deployment update process should be completed on the servers or configuration pages respectively.",
+    "validate.config.files.msg.warning": "Modifying the model directly may result in the cloud being in a bad state.",
```
jeremy-moffitt: I would use different language here. "Modifying the model directly is an advanced use case and should be done with caution as incorrect changes may result in the cloud being in a bad state"

jeremy-moffitt (changes requested):
I'm a little confused, is "prepare deployment" *not* a requirement before deploy? When I run "Prepare deployment" it runs through the configure processor and a couple other steps before succeeding and dropping me back at the file list without validation having completed or the ability to deploy.

So I can run:
`validate->prepare deployment`
`validate->deploy`

but not:
`validate->prepare deployment->deploy`

I understand there is a use case for running `prepare deployment` without `deploy` or as a standalone step, but as a user coming into this screen I'd have expected `<make changes> -> validate -> prepare deployment -> deploy`  based on the buttons shown anyway... so we probably need to either change it so you can run `deploy` after `prepare`... or change the name from `prepare deployment` to something else to indicate that its an alternative to `deploy` and not a prerequisite

jeremy-moffitt:
@guglo per our conversation, I took another look at this and at a minimum we'd need to rename the "prepare deployment" button (in addition to the text feedback above). We might also consider not clearing the flag that allows for deployment at the end of a "prepare deployment" run (unless changes are made afterwards to the model). If thats an option, it might be better than doing the rename. It's definitely confusing that running "Prepare Deployment" disables the "Deploy" button.

guglo:
@jeremy-moffitt  the buttons are disabled based on the existing validation status check, after finish the playbook run and go back the validation page, the validation status becomes unknown state because of the component gets reloaded.  It would be nice that the buttons and validation status check can be coordinated, the model page with buttons are the outer wrapper of the validation component, I need to look into how much of the work to get them coordinate with each other.

jeremy-moffitt:
Maybe instead of 2 buttons we have a checkbox or something to make the deploy button an "either/or" ... since "Deploy" does everything that "Prepare Deployment" does plus a bunch of other things... I'm not sure what we'd call it "Stop after staging"... err... that sounds horrible... "Prepare Only" maintains the current problem that the word "prepare" implies that once its done, the user can do it for real

@mserigithub remind me the use case for only doing the prepare deployment? I think it was to clean-up after an add/replace failure?

guglo:
What I have now is one Update button, click Update button will pop up a dialog, in the dialog there is a checkbox for user to decide if need to run full deployment.

src/localization/bundles/en.json
```diff
@@ -478,6 +480,15 @@
     "deploy.get.log.error": "Failed to get the log for playbook {0} - {1}. {2}",
     "deploy.progress.failure": "Cloud provisioning has failed. Please review the errors in the log, hit Back, address the issue by editing any necessary files, re-run Validate, and Deploy again.",
 
+    "update.progress.heading": "Cloud update deployment in progress",
+    "update.prepare.progress.heading": "Preparing model update deployment in progress",
+    "update.progress.failure": "Failed to deploy cloud update",
+    "update.deploy.confirm": "The deployment of the model update may disrupt your cloud.  Do you want to proceed?",
+    "update.model.heading": "Update model files",
+    "update.model.confirm.title": "Update model",
+    "update.model.confirm.msg": "This will run config-processor-run.yml and ready-deployment.yml playbooks for the model update. Check Run full deployment if want to deploy the model changes to the cloud. \n\nRunning prepare deployemnt playbooks or full deployment playbooks may cause inconsistency of the cloud. Do you want to proceed? ",
```
jeremy-moffitt: I would change the first sentence to:
This will run the config-processor-run.yml and ready-deployment.yml playbooks to update the model.

I would change `Check Run full deployment if want to deploy the model changes to the cloud` to `Check "Run full deployment" to deploy the model changes to the cloud`

there's a misspelling later in the paragraph
`deployemnt` should be `deployment`

src/localization/bundles/en.json
```diff
@@ -478,6 +480,15 @@
     "deploy.get.log.error": "Failed to get the log for playbook {0} - {1}. {2}",
     "deploy.progress.failure": "Cloud provisioning has failed. Please review the errors in the log, hit Back, address the issue by editing any necessary files, re-run Validate, and Deploy again.",
 
+    "update.progress.heading": "Cloud update deployment in progress",
+    "update.prepare.progress.heading": "Preparing model update deployment in progress",
+    "update.progress.failure": "Failed to deploy cloud update",
+    "update.deploy.confirm": "The deployment of the model update may disrupt your cloud.  Do you want to proceed?",
```
jeremy-moffitt: I think the EOS guidelines discourage "you" and "your" ... @dotNomad may be able to confirm... I'd lean towards something like `Deploying a model update may disrupt cloud operations. Proceed with model update?`

jeremy-moffitt (changes requested):
Overall functionality looks good, I think we need to wordsmith a few things

src/components/Modals.js
```diff
@@ -217,3 +217,44 @@ export class SetEncryptKeyModal extends Component {
...
+    event?.preventDefault();
+    this.props.doneAction(this.state.deploy);
+  }
+
+  render() {
+    const footer = (
+      <div className="btn-row">
+        <ActionButton type='default'
+          clickAction={this.props.cancelAction} displayLabel={translate('no')}/>
+        <ActionButton
```
jeremy-moffitt: I would use "Ok" and "Cancel" or "Yes" and "No"... but not "Ok" and "No"....

jeremy-moffitt (approved):

src/localization/bundles/en.json
dotNomad: The guide doesn't explicitly limit "you" and "your" but it does have rules around it. Such as
> Show our expertise by being precise and polite but not personal or overly friendly.

src/components/Modals.js
```diff
@@ -217,3 +217,44 @@ export class SetEncryptKeyModal extends Component {
...
+    this.props.doneAction(this.state.deploy);
+  }
+
+  render() {
+    const footer = (
+      <div className="btn-row">
+        <ActionButton type='default'
+          clickAction={this.props.cancelAction} displayLabel={translate('cancel')}/>
+        <ActionButton
+          clickAction={::this.handleDoneAction} displayLabel={translate('ok')}/>
```
dotNomad: I always find "Ok" to be a bit unclear in menus. I think changing this to also use "Update" may be more clear.

src/localization/bundles/en.json
```diff
@@ -139,7 +140,8 @@
     "validate.config.files.heading": "Review configuration files",
     "validate.config.files.validate": "Validate",
     "validate.config.files.deploy": "Deploy",
-    "validate.config.files.msg.info0": "Validated model changes will be committed but are not automatically deployed. The model may be edited here in support of server add/replace use cases, or as needed to support service configuration changes. The deployment update process should be completed on the servers or configuration pages respectively.",
+    "validate.config.files.msg.warning": "Modifying the model directly is an advanced use case and should be done with caution as incorrect changes may result in the cloud being in a bad state",
```
dotNomad: Need to add a period at the end of the warning.

src/localization/bundles/en.json
```diff
@@ -478,6 +480,15 @@
     "deploy.get.log.error": "Failed to get the log for playbook {0} - {1}. {2}",
     "deploy.progress.failure": "Cloud provisioning has failed. Please review the errors in the log, hit Back, address the issue by editing any necessary files, re-run Validate, and Deploy again.",
 
+    "update.progress.heading": "Cloud update deployment in progress",
+    "update.prepare.progress.heading": "Preparing model update deployment in progress",
+    "update.progress.failure": "Failed to deploy cloud update",
```
dotNomad: Should also include a period at the end of the failure message.

dotNomad (commented):
Functionality looks good to me. Initially thought that checking "Run full deployment" was a little strange, but looking at the comments above understand how that decision was made.
Just had some super small changes with wording and punctuation.

dotNomad (approved):
Looks great :+1:

PR #350, opened 2019-04-16
Add support for running in Docker container (SOC-8524)

GarySmith:


GarySmith:
Need to wait until after cloud 9 ships to merge this

jeremy-moffitt:
whats involved in running/testing this? is there another repo required somewhere? if its just these changes maybe add something to the README?

guglo:
update the jira item to be SOC

GarySmith:
> whats involved in running/testing this? is there another repo required somewhere? if its just these changes maybe add something to the README?

The standard docker commands (e.g. docker up) can be used to bring up the service in a container, and it should work normally there provided you have the ardana service and keystone running. (like normal).

In order to run all three services (ui, ardana service, keystone) in containers and wired up to talk to each other, a new repo, git@github.com:GarySmith/docker-ui.git , was created for this. That repo basically contains the docker compose files and git submodule links to the other three repos. We need to decide whether to try to bring that under the ArdanaCLM namespace in github, and if so, lobby to get access to do so.

jeremy-moffitt:
Was able to run this successfully using docker build and docker run , for the keystone docker the run command doesn't require some random container id, is that possible here too? This definitely works, it may still be worth a couple lines in the readme to help folks along.

GarySmith:
Merged this due to the fact that Jeremy approved it (and I also had tested it), and that none of the files are used in production, dev, or test, but only enables using it in docker

PR #351, opened 2019-04-17
Open useful links in a new tab (SCRD-8910)

dotNomad:


dotNomad:
Documentation on why the target attribute and why I'm using the two rel (relationship) values: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target

PR #353, opened 2019-05-09
Prevent operations when replacing a server in progress (SOC-8756)

ghost:
This commit includes:
1) Disable buttons for adding servers manually or importing servers from csv
when replacing a server in progress.
2) Disable dragging and dropping from manually added servers table or auto
discovered servers table when replacing a server in progress.
3) Don't show dropdown of roles when edit a manually added server during
server replacement.
4) Fixed the property name progressOpertion passed to be consistent with
processOperation.
5) Minor cleanups.
6) Disable the close button in the process page for replacing a controller, activating/deactivating/deleting a compute server.

### DIFF
--- a/src/components/ServerTable.js
+++ b/src/components/ServerTable.js
@@ -26,30 +26,42 @@ class ServerTable extends Component {
       this.props.tableData.map((row, index) => {
         let extraProps = {};
         extraProps.isDraggable = true;
-        // when it is in addserver mode, if we have a list of deployed servers
-        if(this.props.isUpdateMode &&
-           this.props.deployedServers && this.props.deployedServers.length > 0) {
-          // If the server item is NOT in the deployedServers, will present
-          // editAction and deleteAction
-          // Or if isSafeMode is false, will present editAction and deleteAction
-          // And if any other playbook in progress, will not present editActioin
-          // and deleteAction
-          if((!this.props.deployedServers.some(server => {
-            return server['id'] === row['id'] && server['ip-addr'] === row['ip-addr'];
-          }) || !this.props.isSafeMode) && !this.props.progressOperation) {
+        // when it is day2 UI
+        if(this.props.isUpdateMode) {
+          // if we have a list of deployed servers for the server role table
+          if (this.props.id == 'rightTableId' &&
+            this.props.deployedServers && this.props.deployedServers.length > 0) {
+            // If the server item is NOT in the deployedServers, will present
+            // editAction and deleteAction
+            // Or if isSafeMode is false, will present editAction and deleteAction
+            // And if any other playbook in progress, will not present editActioin
+            // and deleteAction
+            if ((!this.props.deployedServers.some(server => {
+              return server['id'] === row['id'] && server['ip-addr'] === row['ip-addr'];
+            }) || !this.props.isSafeMode) && !this.props.processOperation) {
+              extraProps.editAction = this.props.editAction;
+              extraProps.deleteAction = this.props.deleteAction;
+            }
+            // if the server item is in the deployedServers, will NOT present
+            // editAction and deleteAction and item is not draggable
+            else {
+              extraProps.isDraggable = false;
+            }
+
+            // check if any newly added servers have duplicate addresses
+            extraProps.checkNewDupAddresses = this.props.checkNewDupAddresses;
+            extraProps.checkNewDupAddresses['deployedServerIds'] =
+              this.props.deployedServers.map(server => server.id);
+          }
+          // If is the manually added servers or auto discovered servers table, will prevent drag and drop
+          // when process is in progress, for example replacing a server is in progress
+          else if (this.props.id.startsWith('leftTableId')) {
+            if(this.props.processOperation) {
+              extraProps.isDraggable = false;
+            }
             extraProps.editAction = this.props.editAction;
             extraProps.deleteAction = this.props.deleteAction;
           }
-          // if the server item is in the deployedServers, will NOT present
-          // editAction and deleteAction and item is not draggable
-          else {
-            extraProps.isDraggable = false;
-          }
-
-          // check if any newly added servers have duplicate addresses
-          extraProps.checkNewDupAddresses = this.props.checkNewDupAddresses;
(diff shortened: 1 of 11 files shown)


jeremy-moffitt:
@guglo how should I test the "during a server replacement" functionality? whenever I go to the "servers" menu during a server replacement the UI immediately loads the replacement progress

guglo:
@jeremy-moffitt , here are how to check:

In day2,  first go to server page, pick up a controller to replace, enter the mac addr, etc and let it start replace, it goes fast, but don't close the window.  Go to Add Servers, test the followings: 
1. If you don't have anything in the manually added server table, the middle button/link should not be clickable.
2. You can still discover servers, but the the servers discovered should not be draggable. 
3. If have servers in the manually added server table already,  those should not be draggable. Also the buttons for adding and importing are disabled. 
4. If have servers in the manually added server table, click edit, the drop down for changing role is not there. 

5. Go to server summary page, where the page is still show the playbooks steps. close it, then go to add servers page, everything should be tclickable, drggable and dropdown for roles should shop up when edit manually added server. 

In day0, sanity check to make sure no changes in the add server role page.

jeremy-moffitt:
@guglo  ah, I had missed that the "add server" tab was still accessible at the top of the page during a replacement operation... 
I mostly followed the test suggestions above, but I'm a little confused about:
```
Go to server summary page, where the page is still show the playbooks steps. close it, then go to add servers page, everything should be clickable, drggable and dropdown for roles should shop up when edit manually added server.
```
is this basically saying that once the replace is done it should all work again? or is this still during the replacement operation (which seems confusing and contradictory to the statements above it)... guess I'm not sure what is meant by the instruction to "close" the server summary page.

guglo:
@jeremy-moffitt  If replace controller in server summary, there is a button close on the progress page, close that progress page, go to add server page, everything should be back to normal.

jeremy-moffitt:
> @jeremy-moffitt If replace controller in server summary, there is a button close on the progress page, close that progress page, go to add server page, everything should be back to normal.

@guglo you mean the close button at the very end once everything is done, or is there another close button I'm overlooking?

guglo:
image

guglo:
@jeremy-moffitt Sorry for the confusion, please see the screen, that is the close button.

jeremy-moffitt:
@guglo I see now, I was only trying with compute nodes....

this raises an interesting question for me though, does "close" completely cancel the replace? there is a separate "cancel" button on screen, but clicking "close" (at least in the dev env) seemed to have the same effect...

guglo:
@jeremy-moffitt   For multiple process pages in a compute server replacement, we have next on process page, close at the last info mage. cancel and retry (at bottom)  only appear when that specific process page has errors.  The cancel at bottom is different from the cancel in the middle,  cancel in the bottom means that stop here and don't move to the next page. For replacing a controller, it was designed to have one page, there is no  cancel (at bottom), only retry(show up if there are errors) and close. Close means close the page here.

jeremy-moffitt:
scrd-8756-0

guglo:
@jeremy-moffitt  ah...that is a bug, it should be disabled when playbook is in progress and enabled when it is done or playbook failed.

jeremy-moffitt (approved):
functionality works as intended, thanks for fixing the close button issue. Will probably want to get some more eyes on the code before merging in case I missed anything

jeremy-moffitt (approved):
re-tested after the latest commits, still looks good to me

src/pages/ManageComputeHost/ActivateComputeHost.js
```diff
@@ -106,6 +107,13 @@ class ActivateComputeHost extends BaseUpdateWizardPage {
     this.setState(state);
   }
 
+  setCloseButtonDisabled = () => {
```
dotNomad: I see you have the same `setCloseButtonDisabled` set five times across 5 different components. Would it be better to move this up to `BaseUpdateWizardPage` or another higher level component? If this function has to change, it could result in a problem if one of these duplicates is missed.

dotNomad (changes requested):
I know we don't have a formal style guide - and let me know if this is too nitpicky - but I noticed a few spots where there is more than one prop in a line, in complex (multi-line) elements.
Not sure about others, but I find it fairly easy to miss this props when reading through the code.

I highlighted one example in my review.

Additionally found two other spots that might want another look.

src/components/ServerTable.js
```diff
@@ -26,30 +26,42 @@ class ServerTable extends Component {
...
+          if (this.props.id == 'rightTableId' &&
+            this.props.deployedServers && this.props.deployedServers.length > 0) {
+            // If the server item is NOT in the deployedServers, will present
+            // editAction and deleteAction
+            // Or if isSafeMode is false, will present editAction and deleteAction
+            // And if any other playbook in progress, will not present editActioin
+            // and deleteAction
+            if ((!this.props.deployedServers.some(server => {
+              return server['id'] === row['id'] && server['ip-addr'] === row['ip-addr'];
+            }) || !this.props.isSafeMode) && !this.props.processOperation) {
```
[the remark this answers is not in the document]
guglo: Discussed with @dotNomad ,  this block of logics are existing logics, the changes here is to add one more condition to identify the right table so we can differentiate the discovery/manual added servers table (left) from the accordion model table (right) to make they behave differently. Will leave as it is for now.

src/pages/ManageComputeHost/ActivateComputeHost.js
guglo: Discussed with @dotNomad  will make a common function in BaseUpdateWizardPage, child page will still need to overwrite the setCloseButtonDisabled default behavior.

guglo: found a simpler way to just implement in BaseUpdateWizardPage to take care of all the cases.

dotNomad (approved):
:+1: Works as intended. Wasn't able to drag, click buttons, etc while the replace was in progress.

eapodaca (approved):
LGTM

PR #355, opened 2019-05-29
Add doc link to 503 error from Keystone (SOC-3456)

dotNomad:


src/pages/Login.js
```diff
@@ -102,7 +102,20 @@ class LoginPage extends Component {
         } else if (error.status == 403) {
           this.setState({errorMsg: translate('login.unprivileged')});
         } else if (error.status == 503) {
-          this.setState({errorMsg: translate('login.keystone.error')});
+          let docURL = 'https://www.suse.com/documentation/suse-openstack-cloud-9/' +
```
eapodaca: What do you think about adding this url to the locale bundle?

dotNomad: Can definitely do that, just haven't done that in the past. I think that is a probably a better way to go though. :+1:

jeremy-moffitt:
I agree with the suggestion to put the url in the resource bundle. It might also make sense for the link to open in a new tab rather than the current window

PR #357, opened 2019-08-02
Add ipv6 support (SOC-9677)

GarySmith:
Incorporate third-party library, ipaddr.js, to implement various
functions that validate and process IP addresses and CIDRs, which are
more advanced than simple regex matches. Surround literal ipv6 addresses
with brackets with building URLs.

Replace the only instance of subnet+netmask fields with a CIDR field.
This was the "Manage subnet and netmask" button, now renamed to "Manage
network range", which now prompts for a single CIDR field.  This network
range is saved into servers.yml in the input model.  The network range
is still required to be separate ipv4 subnet/netmask fields in
servers.yml, so this two-way conversion is handled by the UI.  In the
future when a single CIDR field is supported in servers.yml, then the
conversion code can be removed.


src/localization/bundles/en.json
```diff
@@ -513,14 +514,16 @@
...
     "input.validator.ipv4address.error": "Please enter a valid IPV4 Address",
-    "input.validator.ipv4address.exist.error": "Address exists, please enter a different IPV4 Address",
-    "input.validator.ipv4address.availservers.exist.error": "Address exists, please select server {0} from Available Servers",
+    "input.validator.ipaddress.exist.error": "Address exists, please enter a different IP Address",
+    "input.validator.ipaddress.availservers.exist.error": "Address exists, please select server {0} from Available Servers",
     "input.validator.macaddress.error": "Please enter a valid MAC Address",
     "input.validator.macaddress.exist.error": "Address exists, please enter a different MAC Address",
     "input.validator.macaddress.availserver.exist.error": "Address exists, please select server {0} from Available Servers",
     "input.validator.port.error": "Please enter a valid port between 0 and 65535",
     "input.validator.ipv4addresshost.error": "Please enter a valid IPV4 Address or Host",
```
guglo: nit: can you remove this? It is not used anymore.

guglo:
@GarySmith  thanks for the changes. I tried to test as I can and found a couple of issues to double check. I used the ipv6 examples in https://www.mediawiki.org/wiki/Help:Range_blocks/IPv6

1. manage cloud settings -> networks add a network in day0

addresses 
2001:d80::
2001:0dbf:ffff:ffff:ffff:ffff:ffff:ffff
2001:d80::-2001:0dbf:ffff:ffff:ffff:ffff:ffff:ffff
2001:db8::/29

sometimes it has errors, sometimes it doesn't. save button doesn't get enabled if enter the addresses above.

Nit: tooltip of cidr or addresses doesn't include a example for ipv6. I am not strong about that. 

2. manage network setttings

a couple of cidr like, it gives error, but save button enabled. 
2001:db8::/29

guglo:
Reading the descriptions of the pr, I don't think I understand clearly, will talk to @GarySmith

GarySmith:
> 1. manage cloud settings -> networks add a network in day0
> 
> addresses
> 2001:d80::
> 2001:0dbf:ffff:ffff:ffff:ffff:ffff:ffff
> 2001:d80::-2001:0dbf:ffff:ffff:ffff:ffff:ffff:ffff
> 2001:db8::/29
I couldn't reproduce this -- all were rejected. I did, however, find that some bogus values like '555555' *are* incorrectly accepted by ipaddr.js as valid ip addresses. I suppose we could file a bug against ipaddr.js for that.

guglo:
After talked to @GarySmith  now I understand addresses in network tab and cidr in cloud network range only accept ipv4 form. 

Figured out how to reproduce the problem for addresses field in network tab

To make error not show up
1. click add network
2. copy and paste 2001:800::/21 in addresses

To make error show up
1. click add network
2. start typing 1.1. in addresses, it has errors
3. copy and paste 2001:800::/21 in addresses

GarySmith:
> 2\. 2001:800::/21

Good catches. Actually this behavior is present in master as well, but I can address it now.

guglo:
This is might not be the issue of our code. If copy and paste directly, it won't complain even 1.1.1. If start typing..it picks it up the changes. My guess is it might not be in our code, it might be something deeper.

GarySmith:
Thanks for the approval. I think we should handle that problem as a separate bug since it is unrelated to this change.

eapodaca:
> LGTM

D'Oh! already merged!

PR #358, opened 2019-08-14
Add server state column (SOC-9957)

eapodaca:


config.dev.js
```diff
@@ -14,7 +14,7 @@
 **/
 
 const config = {
-  'shimurl': 'http://localhost:9085',
+  'shimurl': 'https://192.168.245.6:9085',
```
jeremy-moffitt: guessing this was for testing, probably not ideal for committing

src/pages/UpdateServers.js
```diff
@@ -983,6 +984,22 @@ class UpdateServers extends BaseUpdateWizardPage {
     const autoServers = this.state.servers.filter(s => s.source !== 'manual');
     const manualServers = this.state.servers.filter(s => s.source === 'manual');
 
+    let serverStates = {};
+
+    // servers that are not compute don't have a active/deactive state
+    this.props.model?.get('inputModel')?.get('servers')?.toJS()?.forEach(s => {
+      if(!s.role.includes('COMPUTE')) {
```
jeremy-moffitt: is this how we check for compute roles in other spots? I dont remember what we do on the "add compute" page :-/  @guglo would probably know

eapodaca: we do exactly this on line 262 of this file.

src/localization/bundles/en.json
```diff
@@ -762,5 +765,6 @@
     "table.column.header.server-group" : "Server group",
     "table.column.header.nic-mapping" : "NIC mapping",
     "table.column.header.mac-addr" : "MAC address",
-    "table.column.header.monascaStatus" : "Status"
+    "table.column.header.monascaStatus" : "Status",
+    "table.column.header.serverState": "State"
```
guglo: @eapodaca Now we have status and state, if both of them are there.. could it be confusing? This is more of a question to think through.

src/pages/UpdateServers.js
guglo: we have a util used in the file called isComputeNode, it would be better to use if(!isComputeNode(s)) here. 
at line 262 .it can be updated to filter(s => isComputeNode(s))

PR #359, opened 2019-08-14
Show all packages on package page (SOC-9958)

eapodaca:


jeremy-moffitt:
I think we're opting to document instead that the venv versions match to the upstream versions... closing this for now

